PR #5, opened 2023-03-03
Generic class for trees

pawel-czyz:
This is the basic tree class plus a generic serialization method to nested dictionary. It is easy to use both with Pydantic's `BaseModel` and it's possible to implement custom serialization for e.g., NumPy arrays.


xgluo:
The new commits look good to me!

xgluo (approved):
I believe the changes can be merged.

PR #7, opened 2023-03-09
Add the distance interface

pawel-czyz:
Hi, what do you think about the following interface for distance functions? The added interface and the distance matrix implemented are going to be useful in the tree inference project, but I'd like it to be quite generic.

In particular:
  - Does this generic interface address the use cases in your project?
  - Are the distances between trees symmetric? (I suspect that it isn't always the case, but I can't name an example now).

### DIFF
--- a/src/pyggdrasil/__init__.py
+++ b/src/pyggdrasil/__init__.py
@@ -1,1 +1,7 @@
 """PYggdrasil: Python module for tree inference and analysis."""
+from pyggdrasil.tree import TreeNode
+from pyggdrasil.interface import TreeDistance
+
+import pyggdrasil.tree_inference as tree_inference
+
+__all__ = ["TreeNode", "TreeDistance", "tree_inference"]
--- a/src/pyggdrasil/__init__.py
+++ b/src/pyggdrasil/__init__.py
@@ -1,7 +1,4 @@
 """PYggdrasil: Python module for tree inference and analysis."""
 from pyggdrasil.tree import TreeNode
-from pyggdrasil.interface import TreeDistance
 
-import pyggdrasil.tree_inference as tree_inference
-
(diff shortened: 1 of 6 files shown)


xgluo (commented):
Thank you for adding the distance interface Pawel!

Regarding your questions: 

1. I think the interface looks very good. My only concern is regarding the tree distance matrix. In our use case, we would like to compute a matrix for at the patient level, i.e. the rows and columns of the distance matrix are patients but not trees. Then there are two sub-cases:
- Each patient has a consensus tree, then the current argument with Sequence[_IntegerTreeRoot] is okay. However, constructing a consensus tree is in itself a problem.
- Each patient has a set of trees sampled from the posterior distribution of some phylogenetic methods. These trees should then be weighted equally. For example, if patient 1 has two trees and patient 2 has three trees, then the matrix entry would be the sum of all pairwise distances between the trees divided by 6. In this case, the current implementation of calculate_distance_matrix seems to have a problem. 

2. Upon a quick check, I think all distance measures we consider (Bourque, CASet, DISC, MLTD, MP3 for now) satisfy the symmetry requirements. However, not all of them satisfy triangle inequality, e.g. MLTD. I guess this is not a problem for the implementation.

pawel-czyz:
Hi Xiang, thanks a lot for your comments!

1. Indeed, this function is very generic and not very powerful per se: in the case you consider this function can only be applied to the consensus tree for each patient, but in the case we consider it seems to suffice. The function you mention (calculating the average distances between patients) would be a very different one (and as you mentioned it should probably work with weighted trees. It's up to discussion how to represent them.)
2. Good point, I forgot about the triangle inequality! I added a method for that. (If we know that a metric is symmetric or that triangle inequality holds, we can sometimes speed up computations.)

xgluo (approved):

PR #10, opened 2023-03-22
Add MCMC template for SCITE inference

pawel-czyz:
Hi @gordonkoehn, this is a template for the simplest MCMC, assuming that the false negative and false positive rates are known (I think it'll suffice for now. We'll see whether it's worth to extend this one and investigate posteriors with a larger model, or go to COMPASS). 

I'd suggest you to review it alongside with reading the SCITE paper (and its supplement) – please, point any bugs, places where documentation is cryptic or suggest how the code could be improved :slightly_smiling_face: 

Also, note that the `logprobability_fn` can be anything – for example SCITE's `loglikelihood + logprior`, but also something else. It may be interesting to try something like
`logprobability_fn(tree) = - length of the longest path from the root to any leaf` and see how it works.

### DIFF
--- a/src/pyggdrasil/tree_inference/_mcmc.py
+++ b/src/pyggdrasil/tree_inference/_mcmc.py
@@ -1,0 +1,228 @@
+"""Markov Chain Monte Carlo inference for mutation trees according to the SCITE model.
+
+Note:
+    This implementation assumes that the false positive
+    and false negative rates are known and provided as input.
+"""
+from typing import Callable, Optional
+import jax
+import math
+from jax import random
+import jax.numpy as jnp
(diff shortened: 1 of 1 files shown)


gordonkoehn:
Thank you for the template - think this will be very helpful for me - I'll approve the review once I understand it sufficiently.

gordonkoehn (commented):
Yes, this code skeleton looks very clear to me ! - Thank you for lining it out. Will get onto fleshing it out.

gordonkoehn (approved):
Understood the structure well now. - Will merge this template and open a new pull request to implement it.

PR #11, opened 2023-03-24
Cell Simulation Script

gordonkoehn:
A script to generate test mutation matrices.

Generat random mutation trees possibly with networkX and sample cell attachment to create perfect or optionaly noisy mutation matrices and store tree, perfect and noisy mutation matrix.

### DIFF
--- a/src/pyggdrasil/scripts/cell_simulation.py
+++ b/src/pyggdrasil/scripts/cell_simulation.py
@@ -14,7 +14,53 @@ import networkx as nx
 import numpy as np
 from jax.random import PRNGKeyArray
 
+# TODO: Ask Pawel if this is the way / modify __init__.py
+import pyggdrasil.tree_inference._simulate as sim
 
+
+def create_parser() -> dict:
+    """
+    Parser for required input user.
+
+    Returns:
+        args: dict
+    """
+
+    parser = argparse.ArgumentParser(
+        description="Generate test cell mutation matrices."
+    )
+    parser.add_argument(
+        "--seed",
+        required=False,
+        help="Seed for random generation",
+        type=int,
+        default=42,
+    )
+    parser.add_argument("--outdir", required=True, help="Path to output directory")
+    parser.add_argument("--n_trees", required=True, help="Number of trees", type=int)
+    parser.add_argument("--n_cells", required=True, help="Number of cells", type=int)
+    parser.add_argument(
+        "--n_mutations", required=True, help="Number of mutations", type=int
+    )
+    parser.add_argument(
+        "--alpha", required=True, help="False Negative rate", type=float
+    )
+    parser.add_argument("--beta", required=True, help="False positive rate", type=float)
+    parser.add_argument(
+        "--strategy",
+        required=False,
+        choices=["UNIFORM_INCLUDE_ROOT", "UNIFORM_EXCLUDE_ROOT"],
+        help="Cell Attachment Strategy",
+        default="UNIFORM_INCLUDE_ROOT",
+    )
+
+    args = parser.parse_args()
+    params = vars(args)
+
+    return params
+
+
+# TODO: check with Pawel for requirements of tree topology
 def generate_random_tree(rng: PRNGKeyArray, n_nodes: int) -> np.ndarray:
     """
     Generates a random tree with n nodes.
@@ -26,6 +72,9 @@ def generate_random_tree(rng: PRNGKeyArray, n_nodes: int) -> np.ndarray:
         adj_matrix: np.ndarray
             Note: nodes are here not self-connected
     """
+    # NOTE: opted to not to used networkx random tree generation
+    # as it used numpy randomness which is not compatible with JAX
+
     # Initialize adjacency matrix with zeros
     adj_matrix = np.zeros((n_nodes, n_nodes))
     # Generate random edges for the tree
@@ -63,25 +112,52 @@ def run_sim(params):
         results: dict
             tree, perfect and noisy mutation matrix
     """
+    ############################################################################
+    # Parameters
+    ############################################################################
+    seed = params["seed"]
+    params["n_trees"]
+    n_cells = params["n_cells"]
+    n_mutations = params["n_mutations"]
+    params["alpha"]
+    params["beta"]
+    strategy = params["strategy"]
 
-    ########################################################################################
+    ############################################################################
+    # Random Seeds
+    ############################################################################
+    rng = random.PRNGKey(seed)
+    rng_tree, rng_cell_attachment = random.split(rng, 2)
+
+    ##############################################################################
     # Generate Trees
-    ########################################################################################
+    ##############################################################################
     # used network X to generate random trees and convert to adjacency matrix
-    tree = generate_random_tree(random.PRNGKey(32), 10)
+    tree = generate_random_tree(rng_tree, n_nodes=n_mutations)
+    # if n_trees <=5:
     print_tree(tree)
 
-    ########################################################################################
+    ##############################################################################
     # Attach Cells To Tree
-    ########################################################################################
+    ###############################################################################
+    # convert adjacency matrix to self-connected tree - in tree_inference format
+    np.fill_diagonal(tree, 1)
+    # define strategy
+    strategy = sim.CellAttachmentStrategy[strategy]
+    # attach cells to tree
+    perfect_mutation_mat = sim.attach_cells_to_tree(
+        rng_cell_attachment, tree, n_cells, strategy
+    )
 
-    ########################################################################################
+    print(perfect_mutation_mat)
+
+    ###############################################################################
     # Add Noise
-    ########################################################################################
+    ################################################################################
 
-    #########################################################################################
+    ################################################################################
     # Save Simulation Results
-    #########################################################################################
+    ################################################################################
 
     # Save Tree
     # convert to serialized
@@ -93,40 +169,12 @@ def run_sim(params):
     raise NotImplementedError
 
 
-def create_parser() -> dict:
-    """
-    Parser for required input user.
-
-    Returns:
-        args: dict
-    """
-
-    parser = argparse.ArgumentParser(
-        description="Generate test cell mutation matrices."
-    )
-    parser.add_argument("--outdir", required=True, help="Path to output directory")
-    parser.add_argument("--n_trees", required=True, help="Number of trees", type=int)
-    parser.add_argument("--n_cells", required=True, help="Number of cells", type=int)
-    parser.add_argument(
-        "--n_mutations", required=True, help="Number of mutations", type=int
-    )
-    parser.add_argument(
-        "--alpha", required=True, help="False Negative rate", type=float
-    )
-    parser.add_argument("--beta", required=True, help="False positive rate", type=float)
-
-    args = parser.parse_args()
-    params = vars(args)
-
-    return params
-
-
 def main() -> None:
     """
     Main function.
     """
-    create_parser()
-    # run_sim(params)
+    params = create_parser()
+    run_sim(params)
 
     # with open(file_name, "w") as file_handler:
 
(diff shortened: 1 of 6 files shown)


src/pyggdrasil/scripts/cell_simulation.py
```diff
@@ -0,0 +1,94 @@
...
+    parser.add_argument("--beta", required=True, help="False positive rate", type=float)
+
+    args = parser.parse_args()
+    outdir = args.outdir.strip("/")
+
+    os.makedirs(args.outdir, exist_ok=True)
+
+    # Create the output file
+    file_name = "{}/output.txt".format(args.outdir)
+    try:
```
pawel-czyz, line 86: It's better to use:
```
with open(file_name, "w") as file_handler:
    ...
```

gordonkoehn: Thank you for the suggestion!

src/pyggdrasil/scripts/cell_simulation.py
```diff
@@ -0,0 +1,94 @@
...
+    # Perfect
+    # Noisy
+
+    raise NotImplementedError
+
+
+#########################################################################################
+# MAIN
+########################################################################################
+if __name__ == "__main__":
```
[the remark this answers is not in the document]
gordonkoehn: Yes, will split it up further.

src/pyggdrasil/scripts/cell_simulation.py
```diff
@@ -0,0 +1,186 @@
...
+    # Initialize adjacency matrix with zeros
+    adj_matrix = np.zeros((n_nodes, n_nodes))
+    # Generate random edges for the tree
+    for i in range(1, n_nodes):
+        # Select a random parent node from previously added nodes
+        parent = random.choice(rng, i)
+        # Add an edge from the parent to the current node
+        adj_matrix[parent, i] = 1
+    # Return the adjacency matrix
+    return adj_matrix
```
gordonkoehn, line 87: Check which node is the root. - May have iterated in reverse.

src/pyggdrasil/scripts/cell_simulation.py
```diff
@@ -0,0 +1,369 @@
...
+import argparse
+import jax.random as random
+import numpy as np
+from jax.random import PRNGKeyArray
+import json
+import os
+
+from pyggdrasil.tree import TreeNode
+from pyggdrasil.serialize._to_json import serialize_tree_to_dict
+import pyggdrasil.tree_inference._simulate as sim
```
gordonkoehn, line 20: Is this the correct/pretty way to import these still private functions from a script like this? - Or shall we make them accessible via their respective `__initi__.py`

src/pyggdrasil/scripts/cell_simulation.py
```diff
@@ -0,0 +1,369 @@
...
+
+        # print(f"Parent of node {node} is {child_parent[node]}")
+
+        if node == root_idx:
+            root = TreeNode(name=node, data=None, parent=None)
+            list_TreeNode[node] = root
+        else:
+            # Recall parent
+            parent = child_parent[node]
+            child = TreeNode(name=node, data=None, parent=list_TreeNode[parent])
```
gordonkoehn, line 194: pyright has been giving me a hard time here. This line gives a type error, which I have so far failed to resolve. See the error message below. Any idea?

`list_TreeNode` is an initially empty list of TreeNode elements that is filled with the actual nodes of the tree by transversing the tree using depth-search search. `parent` is just the `int` index of a node in that list.

> cell_simulation.py:194:59 - error: Argument of type "ndarray[Any, dtype[Unknown]]" cannot be assigned to parameter "parent" of type "TreeNode[Unknown, Unknown] | None" in function "__init__"
    Type "ndarray[Any, dtype[Unknown]]" cannot be assigned to type "TreeNode[Unknown, Unknown] | None"
      "ndarray[Any, dtype[Unknown]]" is incompatible with "TreeNode[Unknown, Unknown]"
      Type cannot be assigned to type "None"

gordonkoehn (commented):
First full version of `cell_simulation` - 1 pyright type error to resolve -  annotations and questions to @pawel-czyz

gordonkoehn:
Also, I could not think of any useful unit tests to write for this script. - Any suggestions?

src/pyggdrasil/scripts/cell_simulation.py
```diff
@@ -0,0 +1,398 @@
...
+from jax.random import PRNGKeyArray
+import json
+import os
+
+from pyggdrasil.tree import TreeNode
+from pyggdrasil.serialize._to_json import serialize_tree_to_dict
+import pyggdrasil.tree_inference._simulate as sim
+
+
+def create_parser() -> dict:
```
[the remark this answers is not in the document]
pawel-czyz: Yes, great work! I haven't actually thought about making it interactive (which I have done only once so far).

A stylistic thing for the future: instead of converting `argparse.Namespace` to `dict`, you can just return it and use `args.seed` rather than `dict_of_args["seed"]`.

(Educational comment: feel free to resolve without taking any action.)

src/pyggdrasil/scripts/cell_simulation.py
```diff
@@ -0,0 +1,369 @@
...
+        type=bool,
+    )
+
+    args = parser.parse_args()
+    params = vars(args)
+
+    return params
+
+
+def generate_random_tree(rng: PRNGKeyArray, n_nodes: int) -> np.ndarray:
```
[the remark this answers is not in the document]
pawel-czyz: Sure! :slightly_smiling_face: 

Speaking of reproducibility, JAX solution is really great if you do complex pipelines (with multiple sampling steps) or use parallelization. For scripts, if the sampling is not that complex, using plain NumPy also makes perfect sense.

(Educational comment: feel free to resolve without taking any action.)

src/pyggdrasil/scripts/cell_simulation.py
```diff
@@ -0,0 +1,369 @@
...
+    """
+    Generates a random tree with n nodes, where the root is the first node.
+    Args:
+        rng: JAX random number generator
+        n_nodes: int number of nodes in the tree
+
+    Returns:
+        adj_matrix: np.ndarray
```
pawel-czyz, line 94: I think it'd be good to specify here whether `adj_matrix[i, j]` means an edge "i->j" or the other way around :wink:

src/pyggdrasil/scripts/cell_simulation.py
```diff
@@ -0,0 +1,398 @@
...
+            adjacency matrix
+            Note 1: nodes are here not self-connected
+            Note 2: the root is the first node
+    """
```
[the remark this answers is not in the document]
pawel-czyz: There are different ways of generating the trees (e.g., using evolutionary models), but for our purposes this is perfect.

I actually very much like this function – would you mind moving it to the core library utilities?

src/pyggdrasil/scripts/cell_simulation.py
```diff
@@ -0,0 +1,369 @@
...
+    if (beta > 0) or (alpha > 0) or (na_rate > 0):
+        noisy_mutation_mat = sim.add_noise_to_perfect_matrix(
+            rng_noise, perfect_mutation_mat, alpha, beta, na_rate, observe_homozygous
+        )
+
+    ################################################################################
+    # Save Simulation Results
+    ################################################################################
+    # make save name and path from parameters
+    filename = compose_save_name(params, tree_no=tree_no) + ".json"
```
pawel-czyz, line 289: (Educational comment: feel free to resolve.)

`os` and `os.path` of course work and are a traditional way of working with the filesystem, but Python 3 has a nice new [pathlib library](https://docs.python.org/3/library/pathlib.html) with `pathlib.Path` object. I'm pretty sure you'll like it :wink:

src/pyggdrasil/scripts/cell_simulation.py
pawel-czyz: You've raised a very good point – we  should perfectly use the public API only, exporting the used functions to the `__init__.py` (right now everything except from `TreeNode` seems to live in private submodules).

Generally, the reasoning is as this:
- It's easier to add a function than to remove it from the codebase (so that the users don't have errors when they update the package).
- But if we don't have a public API and use private API anyway, then the purpose of having private API doesn't really exist.

I think we can proceed with adding functions that are necessary in the scripts to the `__init__.py` and the API will form organically. (At some point we'll need to probably trim it down/refactor, but let's not think about it now).

src/pyggdrasil/scripts/cell_simulation.py
pawel-czyz: Ah, I see! `pyright` has a problem with inferring the type of elements stored in the NumPy array, so this is just false positive.

Adding `  # type: ignore` comment to line 194 should silence the error.

src/pyggdrasil/scripts/cell_simulation.py
```diff
@@ -0,0 +1,398 @@
...
+                # Commit Parent to Memory
+                child_parent[child] = node
+
+    root = list_TreeNode[root_idx]
+    return root
+
+
+def dummy_serialize(root: TreeNode):
+    """Dummy function to serialize the data a TreeNode."""
+    pass
```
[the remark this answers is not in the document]
pawel-czyz: This is one way of doing it. As an alternative you can just pass `lambda x: None` and it will work as well :slightly_smiling_face: 

(I was considering making it the default behavior. However, then there's a risk that some data which took hours to generate wouldn't be saved because someone forgot to specify the serialization function here. So passing `lambda x: None` is currently the recommended way of dong it.)

src/pyggdrasil/scripts/cell_simulation.py
```diff
@@ -0,0 +1,369 @@
...
+    Returns:
+        None
+    """
+
+    # Create a random number generator
+    rng = random.PRNGKey(params["seed"])
+
+    # Generate n_simulations of simulated data
+    for i in range(params["n_trees"]):
+        print(f"Generating simulation {i+1}/{params['n_trees']}")
```
pawel-czyz, line 345: Currently `rng` is used twice. The JAX way of doing it is splitting the key before the usage.

In other words, this loop should read:
```python
 for i in range(params["n_trees"]):
        print(f"Generating simulation {i+1}/{params['n_trees']}")
         rng, key_iter = random.split(rng)
        gen_sim_data(params, key_iter, tree_no=i + 1)
```

pawel-czyz: There's also another way, which I actually like the most, with generating different random keys for different iterations at the start:

```python
# Create a random number generator
rng = random.PRNGKey(params["seed"])
keys = random.split(rng, params["n_trees"])
for i, key in enumerate(keys, 1):
    print(f"Generating simulation {i}/{len(keys)}")
    gen_sim_data(params, key, tree_no=i)
```

pawel-czyz (approved):
Very nice PR!

I think two things would be necessary to merge this PR:
1. Silencing the false positive type error (see below).
2. Fixing an issue with the random keys (see below).

The rest are just stylistic suggestions, which are not really necessary to address, but perhaps you'll find them useful in your future work.

Good work, keep going! :slightly_smiling_face:

src/pyggdrasil/scripts/cell_simulation.py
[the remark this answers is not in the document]
gordonkoehn: Yes, moved it.

src/pyggdrasil/scripts/cell_simulation.py
gordonkoehn: Will have a look for next time.

src/pyggdrasil/scripts/cell_simulation.py
gordonkoehn: Oh, yes -  good catch -  implemented your second proposal.

PR #13, opened 2023-03-30
Move scripts outside of the package

pawel-czyz:
Accidentally `scripts` became a subpackage, rather than living "next to it" in the repository tree. This PR corrects the structure.


gordonkoehn (approved):
Oh yes - my mistake. Thanks for correcting it!

PR #16, opened 2023-04-14
Adding Workflow Management

gordonkoehn:
Implementation of reproducible and scalable workflows using `snakemake`. Currently, only implements the mock data generation using `cell_simulation.py`.

The code structure and environmental setup are described in the `workflows/README.md`

Looking forward to your feedback and thoughts on the setup!

### DIFF
--- a/scripts/cell_simulation.py
+++ b/scripts/cell_simulation.py
@@ -6,9 +6,18 @@ by generating random trees and sampling cell attachments
 with noise if error rates are non-zero.
 
 As per definition in the SCITE Jahn et al. 2016.
+
+Example Usage:
+poetry run python ../scripts/cell_simulation.py
+--seed 42
+--out_dir ../data --n_trees 3 --n_cells 100 --n_mutations 8
+ --strategy UNIFORM_INCLUDE_ROOT --alpha 0.01 --beta 0.02
+ --na_rate 0.01 --observe_homozygous False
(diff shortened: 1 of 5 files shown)


scripts/cell_simulation.py
```diff
@@ -67,7 +94,8 @@ def create_parser() -> argparse.Namespace:
         "--observe_homozygous",
         required=True,
         help="Observing homozygous mutations",
-        type=bool,
+        choices=[False, True],
+        type=t_or_f,
```
gordonkoehn, line 98: There was a bug here. `cell_simulation.py` did not parse a string "False" correctly. It always parsed `True`. Hence, this custom function `t_or_f`. Please let me know if there is a more elegant way.

pawel-czyz: Actually, I'd recommend not making it required and setting it to `False` by default.

This should work:
```
parser.add_argument(
    "--observe-homozygous",
    help="Observing homozygous mutations. By default False.",
    action="store_true",
)
```

pawel-czyz:
Very nice! Have you had prior experience with Snakemake? It's cool that you know how to use `expand` and wildcards :slightly_smiling_face:

gordonkoehn:
Great. Thanks for the Review! - hope to learn Snakemake properly further down the line.

Issue #21, opened 2023-04-19
Tree visualisation


pawel-czyz:
For a given tree, save an image with it rendered with log-prob in the title.

Sprint: #18

gordonkoehn:
Defining the Product: 

plotTree(Args): that takes a tree and print options to save a report grade visualisation.

Args: 
	- TreeNode or root of Tree
		- Name:  title of the tree (title of the Tree given by the root name root)
		- Data: dict(of arguments to print, on each node)
	- print options
		- print title: name of the root
		- print tree_data :  dict( strings of argument of data to print)
		- print nodes print data : dict(strings of arguments of data to print)
	- save name
	- save dir

This implementation will use the GraphViz tools. Dark blue solid nodes with white labels are imagined. 

[TreeVisualization_Product_2023-04-24.pdf](https://github.com/cbg-ethz/PYggdrasil/files/11309007/TreeVisualization_Product_2023-04-24.pdf)

gordonkoehn:
Dropping the feature to `print node_data` other than the root. 

Reason: no obvious reason for this project. - no facial implementation.

Issue #18, opened 2023-04-19
🏃 Sprint 1: MCMC finding the trees and plotting convergence


pawel-czyz:
This scopes the first sprint, lasting three weeks (**April 19th–May 10th**).

---
As a reader of your thesis, I'd like to see:
  - Some of the true generated trees, to present how mutation trees look like.
  - Some of the sampled trees, when the chain has converged. We want to be able to see the loglikelihood/logposterior of the trees:
![Image](https://user-images.githubusercontent.com/26385111/233063208-792c42e5-1f3e-41f7-a89e-8545f10bb5dc.png)
  - The plot with X axis being iteration number, Y axis being the loglikelihood of the sampled tree.
  - The sampling times.
  - The number of iterations needed to sample the true tree.
  - Plot of the distance to the true tree on the Y axis for several chains and the iteration number on X axis.

We want to have this:
  - For FP and FN rates taken from SCITE simulation paper (probably 1% and 10%, but I don't remember) and 5, 7, 10 mutations. 
  - For a misspecified model, where the true rates are slightly larger that the specified ones to the model used in the MCMC inference. (Example: rather than having 1% and 10% error rates as given to the MCMC SCITE model, we simulate the data from the 2%, 15%).

PR #27, opened 2023-04-24
Tree Visualisation

gordonkoehn:
For a given tree, save an image with it rendered with log-prob in the title.

Sprint:  1) MCMC finding the trees and plotting convergence #18
Issue: Tree visualisation #21


gordonkoehn:
Customizable plot version bases on `networkX` working. 

Allows to pass and turn on/off all print elements, all default latex font. See this:

unit_test_tree


I ended up using `matplotlib` and `netwokX` instead of `dot` and `graphviz` as the latter just lags / I failed to find the capability to convert and anytree AND modify it after. A long struggle.

My only concern  is with the `anytree` to `networkX` conversion. I am making use of the `pydot` library which gives 

` DeprecationWarning: nx.nx_pydot.from_pydot depends on the pydot package, which has known issues and is not actively maintained.
` see Issue https://github.com/networkx/networkx/issues/5723

Perhaps I can see to find another conversion later.

What do you say? - Do we need more features?  - Is this a report-grade figure?

src/pyggdrasil/visualize/tree.py
```diff
@@ -0,0 +1,118 @@
...
+
+    # make title
+    if print_options["title"]:
+        fig.suptitle(tree.data["tree-name"], fontsize=20)
+
+    description = ""
+    for detail in print_options["data_tree"]:
+        if detail == "log-likelihood":
+            description += (
+                r"$\log(P(D|T,\theta)$: " + str(tree.data["log-likelihood"]) + "\n"
```
pawel-czyz, line 110: I think one ")" is missing here.

Also, you can consider using `f"{tree.data['log-likelihood']:.2f}"` (rather than `str(...)`) to control the precision of the printed out number.

tests/visualize/test_viz_tree.py
```diff
@@ -0,0 +1,62 @@
...
+
+from pyggdrasil.tree_inference._tree import Tree
+
+import pyggdrasil.visualize.tree as viz
+
+IN_GITHUB_ACTIONS = os.getenv("GITHUB_ACTIONS") == "true"
+
+
+@pytest.mark.skipif(IN_GITHUB_ACTIONS, reason="Test doesn't work in Github Actions.")
+def test_plot():
```
pawel-czyz, line 18: I'd suggest to use [the `tmp_path` fixture](https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html) to specify the path, rather than using `../data/trees/`. It's also possible (although I don't know how likely) that this was the issue with Github Actions (not all directories there are writeable, if I remember correctly)

pawel-czyz:
> Customizable plot version bases on `networkX` working. I decided to make this a module, instead of a script. I may add scripts using this module later in the workflow.
> 
> Allows to pass and turn on/off all print elements, all default latex font. See this:
> 
> unit_test_tree
> 
> I ended up using `matplotlib` and `netwokX` instead of `dot` and `graphviz` as the latter just lags / I failed to find the capability to convert and anytree AND modify it after. A long struggle.
> 
> My only concern is with the `anytree` to `networkX` conversion. I am making use of the `pydot` library which gives
> 
> `DeprecationWarning: nx.nx_pydot.from_pydot depends on the pydot package, which has known issues and is not actively maintained.` see Issue [networkx/networkx#5723](https://github.com/networkx/networkx/issues/5723)
> 
> Perhaps I can see to find another conversion later.
> 
> What do you say? - Do we need more features? - Is this a report-grade figure?

Very nice! I added a comment regarding the typo and node indexing, but otherwise it looks good to me! :slightly_smiling_face:

src/pyggdrasil/visualize/tree.py
```diff
@@ -0,0 +1,118 @@
...
+    # make output directory if it doesn't exist
+    os.makedirs(save_dir, exist_ok=True)
+
+    # convert to networkX graph
+    # convert to dot graph
+    dot_string = ""
+    for line in DotExporter(tree):
+        dot_string += line
+    # convert to networkX graph
+    graphs = pydot.graph_from_dot_data(dot_string)
```
pawel-czyz, line 53: Is it possible to specify here node relabeling? We may want to (a) enumerate mutations starting from 1, rather than 0 or (b) use biological mutation names.

(If it's too hard for a quick fix, we can also address it later.)

pawel-czyz (approved):

src/pyggdrasil/visualize/tree.py
gordonkoehn: Oh, yes of course. Thanks for spotting.

src/pyggdrasil/visualize/tree.py
gordonkoehn: All implemented. - see the comment below.

Issue #20, opened 2023-04-19
Full MCMC simulation


pawel-czyz:
For a given set of error rates, initial tree, and seed, and cell data, collect the samples and save them to the disk.

In sprint #18

pawel-czyz:
Time estimates: M, L.

gordonkoehn:
Raise value error if nan or homozygous mutation matrix is given.

gordonkoehn:
Add proper formating for runtime and runtime for sample.

gordonkoehn:
Change default read data to `perfect_mutation_mat`

gordonkoehn:
- [ ] consider using `xarray.save_mfdataset` instead of `json` to write mcmc samples.

gordonkoehn:
- [ ] print config file to log.

PR #28, opened 2023-04-27
Full MCMC simulation

gordonkoehn:
For a given set of error rates, initial tree, and seed, and cell data, collect the samples and save them to the disk.

Sprint 1: MCMC finding the trees and plotting convergence #18 
Issue: Full MCMC simulation #20


src/pyggdrasil/tree_inference/_mcmc.py
```diff
@@ -226,8 +228,10 @@ def _validate_move_probabilities(move_probabilities: MoveProbabilities, /) -> No
 def _mcmc_kernel(
     key: random.PRNGKeyArray,
     tree: Tree,
+    data: MutationMatrix,
+    theta: ErrorRates,
     move_probabilities: MoveProbabilities,
-    logprobability_fn: Callable[[Tree], float],
+    logprobability_fn: Callable[[MutationMatrix, Tree, ErrorRates], float],
```
gordonkoehn, line 234: @pawel-czyz I realized that the implementation of the mcmc kernel, initially sketched out by you, feature a `log_probability` function that does not take the `data: Mutation matrix`nor  the `theta:ErrorRates`. Was this by design ? - Currently my `log_prob.logprobability_fn()` requires these arguments, hence these changes here. Is there a smarter way to pass these - as they don't change through the run - perhaps a rare case for a`global` variables.

tests/visualize/test_viz_tree.py
```diff
@@ -14,6 +14,7 @@
 IN_GITHUB_ACTIONS = os.getenv("GITHUB_ACTIONS") == "true"
 
 
+@pytest.mark.filterwarnings("ignore::DeprecationWarning:")
```
gordonkoehn, line 17: Strictly not part of this PR -  part of #27. - ignoring this warning, as we are not heavily dependent on it.

tests/tree_inference/test_mcmc.py
```diff
@@ -191,3 +191,6 @@ def test_swap_subtrees_move():
 
     assert jnp.array_equal(new_tree.tree_topology, new_tree_corr.tree_topology)
     assert jnp.array_equal(new_tree.labels, new_tree_corr.labels)
+
+
+# TODO: test MCMC Kernel - any suggestions?
```
gordonkoehn, line 196: @pawel-czyz Given the `log_prob` is not implemented, I could code some tests for the yet-to-be-tested `mcmc_kernel`. Any ideas?

gordonkoehn (commented):
Rough code skeleton. - Trying to use `jax.lax.while_loop` directly. Got two doubts regarding: 

1. where the read and save functionality of the mcmc samples belong

2. how to pass the observed data and error rates through the `mcmc_kernel` to the `log_probability_fn`

src/pyggdrasil/tree_inference/_mcmc.py
pawel-czyz: I think both options are fine. Essentially, I sketched it having a [curried](https://en.wikipedia.org/wiki/Currying) version of the function in mind:
you can have
```python

def logprob_all(data, tree, rates) -> float:
    return some_expression(...)
```
or:

```python

def create_logprob(data, rates) -> Callable:
    def logprob_(tree) -> float:
        return logprob_all(data, tree, rates)
    return logprob_
```

Currying is a common trick in functional programming and BlackJAX employs it often, but it doesn't mean that we have to (feel free to change the sketch :slightly_smiling_face: )

src/pyggdrasil/serialize/_to_json.py
```diff
@@ -91,3 +92,31 @@ def generate_node(node_dict: dict, parent: Optional[TreeNode]) -> TreeNode:
...
+    """Reads MCMC sample from JSON file.
+
+    Args:
+        output_dir: directory to read sample from
+        sample_id: sample number to read
+
+    Returns:
+        MCMC sample"""
+    # TODO: implement read mcmc sample
+    raise NotImplementedError
```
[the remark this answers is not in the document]
pawel-czyz: I think I'd suggest to keep them here for now. The point is that serialization subpackage doesn't know about the internal format for storing the trees (via matrices).

tests/tree_inference/test_mcmc.py
pawel-czyz: I think we actually have it – isn't log-prob just the log-likelihood function in our case? We have (a) fixed rates, (b) uniform prior on the trees, (c) uniform attachments, so the difference between these two is the log of the normalization constant.

src/pyggdrasil/serialize/_to_json.py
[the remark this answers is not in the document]
gordonkoehn: OK - I was thinking of passing /returning an `xarray` with info already in basic types, not `tree_inference._tree` objects. Another function in `mcmc_util` could then do the conversion, to the xarrays used in the `tree_inference._mcmc_sampler` for a restart/saving of mcmc sampeles.

tests/tree_inference/test_mcmc.py
gordonkoehn: Yes, indeed, not many additional workings in the `mcmc_kernel`. I will skip a test here then. Thanks for double-checking this with me.

scripts/cell_simulation.py
```diff
@@ -164,8 +179,8 @@ def gen_sim_data(
     ##############################################################################
     # Generate Trees
     ##############################################################################
-    #  generate random trees (uniform sampling) as adjacency matrix
-    tree = tree_inf.generate_random_tree(rng_tree, n_nodes=n_mutations)
+    #  generate random trees (uniform sampling) as adjacency matrix / add +1 for root
+    tree = tree_inf.generate_random_tree(rng_tree, n_nodes=n_mutations + 1)
```
gordonkoehn, line 183: Found a bug! - the `cell_simulation.py` did generate random trees with one mutation to little.

scripts/cell_simulation.py
```diff
@@ -28,6 +28,21 @@
...
+        if isinstance(obj, np.integer):
+            return int(obj)
+        if isinstance(obj, np.floating):
+            # 👇️ alternatively use str()
+            return float(obj)
+        if isinstance(obj, np.ndarray):
+            return obj.tolist()
+        return json.JSONEncoder.default(self, obj)
+
+
```
gordonkoehn, line 45: The JSON writer broke down. - don't know why. Might have changed the python version locally. "TypeError: Object of type int64 is not JSON serializable" Implemented this workaround from the web.

gordonkoehn:
@pawel-czyz As you can see from the above two comments there were some issues with the `cell_simulation.py`. I tested this script only manually. Is there a way to use `pytest`s for more rigorous testing of the `scripts/*`. I would like to check for file creation and output shapes.

tests/scripts/test_cell_simulation.py
```diff
@@ -0,0 +1,62 @@
+"""Tests the cell_simulation script."""
```
gordonkoehn, line 1: Wrote a test for the `cell_simulation.py` script here, as I spotted a bug and some errors before. As discussed this `pytest` tests a script directly. - Consider refactoring the logic from the `cell_simulation.py` script to `tree_inference.simulate`.

scripts/cell_simulation.py
pawel-czyz: Right, good you found it!

src/pyggdrasil/tree_inference/_mcmc.py
```diff
@@ -284,6 +290,13 @@ def _mcmc_kernel(
     else:
         proposal, log_q_diff = _swap_subtrees_proposal(key, tree)
 
+    # TODO:Consider replacing the above condition with
+    # condlist = [move_type == 0, move_type == 1]
+    # choicelist = [_prune_and_reattach_proposal(key, tree),
+    #               _swap_node_labels_proposal(key, tree)]
+    # default = _swap_subtrees_proposal(key, tree)
+    # proposal, log_q_diff = jnp.select(condlist, choicelist, default)
+
```
gordonkoehn, line 299: Doubt: In line-comment, I have a `jit` compatible version of these `if, elif, else` conditions. This is a relict from my attempts to use use `jax.lax.while_loop` in the main mcmc loop.  I quickly realized that a bunch of modifications are needed, deeper down in 'get_descendants' and so on. - Fixing this to use jax properly will be in the next sprint I suppose?

src/pyggdrasil/tree_inference/_mcmc_sampler.py
```diff
@@ -0,0 +1,164 @@
...
+    # TODO: use jax.lax.while_loop instead of while_loop,
+    # requires further modification of _get_descendants at least
+    # jax.lax.while_loop(cond_fn, body, init_state)
+
+    def while_loop(cond_fun, body_fun, init_val):
+        """While loop for MCMC sampler."""
+        val = init_val
+        while cond_fun(val):
+            val = body_fun(val)
+        return val
```
gordonkoehn, line 160: Wrote this function in a way, to be usable with `jax.lax.while_loop`. 

Some modifications are needed in the `mcmc kernel` and in deeper down in `get_descendants`. Might need some help there. - Couldn't make sense of some `jax._src.errors.ConcretizationTypeError`.

src/pyggdrasil/tree_inference/_mcmc_util.py
```diff
@@ -96,3 +99,57 @@ def _prune_and_reattach_move(tree: Tree, *, pruned_node: int, attach_to: int) ->
...
+        logprobability : float - log probability of tree
+        rng_key_run : JAXRandomKey - random key used to run MCMC
+    """
+    iteration = ds["iteration"].item()
+    tree = Tree(
+        jnp.array(ds["tree"].values), jnp.array(ds["tree"].coords["from_node_k"].values)
+    )
+    logprobability = ds["log-probability"].item()
+
+    return iteration, tree, logprobability
```
gordonkoehn, line 155: Info: Functionality to pack and unpack mcmc samples, to be easily modifiable.

tests/visualize/test_viz_tree.py
gordonkoehn, line 17: Info: ignored warnings related to `pydot`.

workflows/Snakefile
```diff
@@ -80,5 +83,47 @@ rule generate_mock_data:
...
+    params:
+        out_dir = "../data/mcmc/mark00",
+    shell:
+        """
+        poetry run python {input.script} \
+            --data_fp {input.data} \
+            --config_fp {input.config_file} \
+            --out_dir {params.out_dir}
+        """
+
```
gordonkoehn, line 127: Info: simple rule to run 1 mcmc as specified in the `config.json`

workflows/config/mcmc_config_mark00.json
```diff
@@ -0,0 +1,12 @@
...
+        "prune_and_reattach": 0.1,
+        "swap_node_labels": 0.65,
+        "swap_subtrees": 0.25
+    },
+    "fnr": 0.097,
+    "fpr": 1.24e-06,
+    "num_samples": 100,
+    "burn_in": 10,
+    "thinning": 2
+}
```
gordonkoehn, line 12: I decided to use a config file for `run_mcmc.py`to pass all these arguments as I don't expect them to change that much and it would be cumbersome to pass so many arguments each time. What do you think?

scripts/cell_simulation.py
```diff
@@ -146,92 +146,28 @@ def gen_sim_data(
...
-    ###############################################################################
-    # Add Noise
-    ################################################################################
-    # add noise to perfect mutation matrix
-    noisy_mutation_mat = None
-    if (beta > 0) or (alpha > 0) or (na_rate > 0):
-        noisy_mutation_mat = tree_inf.add_noise_to_perfect_matrix(
-            rng_noise, perfect_mutation_mat, alpha, beta, na_rate, observe_homozygous
-        )
+    data = tree_inf.gen_sim_data(params_dict, rng)
```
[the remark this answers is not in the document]
pawel-czyz: I fully support this idea, I like small scripts :sweat_smile:

scripts/cell_simulation.py
```diff
@@ -146,92 +146,28 @@ def gen_sim_data(
...
 
     ################################################################################
     # Save Simulation Results
     ################################################################################
     # make save name and path from parameters
     filename = compose_save_name(params, tree_no=tree_no) + ".json"
-    fullpath = os.path.join(out_dir, filename)
+    fullpath = os.path.join(params.out_dir, filename)
```
pawel-czyz, line 161: Nothing to change in this PR, but in the future I'd suggest to take a look at [pathlib](https://docs.python.org/3/library/pathlib.html). Perhaps you'll find it more convenient than `os` and `os.path`.

scripts/run_mcmc.py
```diff
@@ -0,0 +1,295 @@
...
+import pyggdrasil.serialize as serialize
+
+from pyggdrasil.tree_inference import (
+    MutationMatrix,
+    mcmc_sampler,
+    MoveProbabilities,  # type: ignore
+)
+
+
+def create_parser() -> argparse.Namespace:
```
pawel-czyz, line 64: Very nice, I see a lot of progress in writing parsers on your side!

scripts/run_mcmc.py
```diff
@@ -0,0 +1,295 @@
...
+        data_fp: str
+            Filepath of the data.
+
+    Returns:
+        mut_mat: MutationMatrix
+            observed mutation matrix
+    """
+
+    # load data from file to json object
+    with open(data_fp, "r") as f:
```
pawel-czyz, line 135: I personally like `"r"` mode here, but it's the default option (so you can also skip it in the future).

scripts/cell_simulation.py
pawel-czyz: Ah, you already know it! I saw that you used it in another file. Sorry for noise!

scripts/run_mcmc.py
```diff
@@ -0,0 +1,295 @@
...
+        move_probs=move_probs,
+        num_samples=config["num_samples"],
+        num_burn_in=config["burn_in"],
+        output_dir=Path(params.out_dir),
+        thinning=config["thinning"],
+        init_tree=init_tree,
+        timestamp=kwargs.get("timestamp", None),
+    )
+
+    print("Done!")
```
pawel-czyz, line 226: Not a problem really, but above you use `logging.info` (which I think should be also preferred in this case).

scripts/run_mcmc.py
```diff
@@ -0,0 +1,295 @@
...
+
+    # convert json object to mutation matrix
+    mut_mat = data["noisy_mutation_mat"]  # TODO: adjust to flexible title
+    # convert to array
+    mut_mat = jnp.array(mut_mat)
+
+    return mut_mat
+
+
+def run_chain(params: argparse.Namespace, config: dict, **kwargs) -> None:
```
pawel-czyz, line 146: For `config` it may be beneficial in the long run to make it a Pydantic object (my favourite solution), a data class, or a [TypedDict](https://peps.python.org/pep-0589/).

pawel-czyz, line 146: Are `**kwargs` necessary here? I may have overlooked something, but I don't think they are used here... 
(And passing weakly-typed arguments is often not the best idea possible.)

scripts/run_mcmc.py
```diff
@@ -0,0 +1,295 @@
...
+            _, init_tree, _ = tree_inf.unpack_sample(mcmc_sample)
+            logging.info("Loaded tree (mcmc sample) from file.")
+        else:
+            raise ValueError(
+                "Please provide either TreeNode or mcmc sample number,"
+                " to read in tree from file."
+            )
+
+    # Make Move Probabilities
+    prune_and_reattach = config["move_probs"]["prune_and_reattach"]
```
pawel-czyz, line 207: Nice that you have it! I think we'll actually use this option quite a lot in the next sprint :slightly_smiling_face:

scripts/run_mcmc.py
```diff
@@ -0,0 +1,295 @@
...
+
+
+def main() -> None:
+    """
+    Main function.
+    """
+    # Parse command line arguments
+    params = create_parser()
+
+    # TODO: check params
```
pawel-czyz, line 253: This is one of the reasons why I love [Pydantic](https://docs.pydantic.dev/latest/) :slightly_smiling_face:

src/pyggdrasil/interface/__init__.py
```diff
@@ -6,3 +6,20 @@
   - This subpackage should not depend on other subpackages, so we
     do not introduce circular imports.
 """
+
+import xarray as xr
+
+
+# MCMC sample in xarray format.
```
pawel-czyz, line 13: Nice!

src/pyggdrasil/serialize/_to_json.py
```diff
@@ -91,3 +113,78 @@ def generate_node(node_dict: dict, parent: Optional[TreeNode]) -> TreeNode:
...
+    Args:
+        output_fp: directory to save tree to
+    Returns:
+        None
+    """
+
+    with open(output_fp, "r") as f:
+        tree_node = json.load(f)
+
+    return deserialize_tree_from_dict(tree_node, deserialize_data=lambda x: x)
```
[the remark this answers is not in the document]
pawel-czyz: Right, makes sense to me.

src/pyggdrasil/serialize/_to_json.py
```diff
@@ -91,3 +113,78 @@ def generate_node(node_dict: dict, parent: Optional[TreeNode]) -> TreeNode:
...
+
+
+def save_mcmc_sample(sample: MCMCSample, output_dir: Path, **kwargs) -> None:
```
pawel-czyz, line 149: Why can't `**kwargs` be replaced with a simple `timestamp: Optional[str] = None`?

src/pyggdrasil/tree_inference/_interface.py
```diff
@@ -50,3 +50,7 @@
 # i.e. [0,0,...0,0,1]
 # A node is its own ancestor.
 AncestorMatrix = Array
+
+# Observational Error rates
+# tuple of (alpha, beta)
+ErrorRates = Tuple[float, float]
```
pawel-czyz, line 56: In the future (when e.g., Euler's Python is sufficiently modern. Or perhaps it's already?), you won't need `Tuple` and ordinary `tuple[float, float]` will work (similarly for `list` and `dict`).

src/pyggdrasil/tree_inference/_logprob.py
```diff
@@ -5,23 +5,53 @@
...
+        Returns:
+            log-probability of the tree
+        """
+        return logprobability_fn(data, tree, rates)
+
+    return logprob_
+
+
+def logprobability_fn(data: MutationMatrix, tree: tr.Tree, theta: ErrorRates) -> float:
+    """Calculates the log-probability of a tree given error rates and data.
```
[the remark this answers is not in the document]
pawel-czyz: Nice!

src/pyggdrasil/tree_inference/_mcmc.py
```diff
@@ -184,7 +186,9 @@ def _swap_subtrees_proposal(key: random.PRNGKeyArray, tree: Tree) -> tuple[Tree,
         desc_node2 = tr._get_descendants(tree.tree_topology, tree.labels, node2)
         # \Delta q = log q(new|old) - log q(old|new) = log [d(i)+1] - log [d(k)+1]
         # where k is the descendant node of i
-        corr = float(jnp.log(desc_node1 + 1.0) - jnp.log(desc_node2 + 1.0))
+        corr = float(
+            jnp.log(desc_node1.shape[0] + 1.0) - jnp.log(desc_node2.shape[0] + 1.0)
```
pawel-czyz, line 190: Here it doesn't matter (as shapes are integers), but perhaps in some of your future projects you'll find this function useful: https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.log1p.html
(It's primary use case is for `log(1+p)` for small `p`).

src/pyggdrasil/tree_inference/_mcmc.py
pawel-czyz: For if/else I often use [jax.lax.cond](https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.cond.html). But don't worry about making it JAX-compatible now. I think with the current implementation we can already do some cool research :wink:

src/pyggdrasil/tree_inference/_mcmc_sampler.py
```diff
@@ -0,0 +1,164 @@
...
+    init_tree: Tree,
+    error_rates: ErrorRates,
+    move_probs: MoveProbabilities,
+    data: MutationMatrix,
+    num_samples: int,
+    output_dir: Path,
+    num_burn_in: int = 0,
+    thinning: int = 0,
+    iteration: int = 0,
+    **kwargs,
```
pawel-czyz, line 42: How much are `**kwargs` used below? From a quick look I think `timestamp: Optional[str] = None` would be a better choice here. (I can also be wrong.)

src/pyggdrasil/tree_inference/_mcmc_sampler.py
```diff
@@ -0,0 +1,164 @@
...
+        tree, logprobability = mcmc._mcmc_kernel(
+            rng_key_sample,
+            tree,
+            data,
+            error_rates,
+            move_probs,
+            logprobability_fn,
+            logprobability,
+        )
+        logging.info("Iteration: %d, log-probability: %f", iter_sample, logprobability)
```
pawel-czyz, line 123: This is one way of doing it, another one is via format strings (and then you can specify e.g., precision in which `float` is reported.)

src/pyggdrasil/tree_inference/_mcmc_sampler.py
pawel-czyz: I think it currently has the problem of tracing the shapes of arrays. One reason for this could be if e.g., number of mutations (as an integer) is passed in the function argument.

src/pyggdrasil/tree_inference/_simulate.py
```diff
@@ -511,8 +512,11 @@ def adjacency_to_root_dfs(
     # Sanity checks
     if adj_matrix.shape[0] != adj_matrix.shape[1]:
         raise ValueError("Adjacency matrix must be square")
-    if adj_matrix.shape[0] != len(labels):
-        raise ValueError("Number of labels must match number of nodes")
+
+    if labels is not None:
+        if adj_matrix.shape[0] != len(labels):
+            raise ValueError("Number of labels must match number of nodes")
+
```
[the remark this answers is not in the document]
pawel-czyz: Good you found it! But how is possible that `labels` can be `None`?

src/pyggdrasil/tree_inference/_simulate.py
```diff
@@ -581,3 +585,98 @@ def adjacency_to_root_dfs(
...
+    Returns:
+        data: dict
+            data dictionary containing - serialised data for the tree:
+                adjacency_matrix - adjacency matrix of the tree
+                perfect_mutation_mat - perfect mutation matrix
+                noisy_mutation_mat - noisy mutation matrix
+                                    (only if alpha > 0 | beta > 0 | na_rate > 0)
+                root - root of the tree (TreeNode)
+
+    """
```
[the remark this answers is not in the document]
pawel-czyz: Good idea :slightly_smiling_face:

src/pyggdrasil/tree_inference/_simulate.py
```diff
@@ -581,3 +585,98 @@ def adjacency_to_root_dfs(
     root = list_tree_node[root_idx]
 
     return root
+
+
+def gen_sim_data(
+    params: dict,
```
pawel-czyz, line 591: (Just repeating myself: I think Pydantic's `BaseModel` is the perfect candidate in use cases like this one.)

tests/tree_inference/test_simulate.py
```diff
@@ -447,3 +447,32 @@ def test_attach_cells_to_tree_case1(
...
+    # check that the dimensions of the data are correct
+    assert np.array(data["adjacency_matrix"]).shape == (8 + 1, 8 + 1)
+    assert np.array(data["noisy_mutation_mat"]).shape == (
+        8 + 1,
+        100,
+    )  # TODO: to be altered if we truncate the matrix
+    assert np.array(data["perfect_mutation_mat"]).shape == (
+        8 + 1,
+        100,
+    )  # TODO: to be altered if we truncate the matrix
```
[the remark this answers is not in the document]
pawel-czyz: Very good idea!

workflows/config/mcmc_config_mark00.json
pawel-czyz: There are two ways how this is often done:

1. Pydantic's `BaseModel` can have default values and have hierarchical configuration.
2. [Hydra](https://hydra.cc/docs/intro/) is Facebook AI Research's solution of maintaining hierarchical configurations.

We should consider perhaps using one of these, but it's definitely out of scope for this PR :slightly_smiling_face:

pawel-czyz (approved):
Very nice! I think there's quite a few new tricks (logging, xarray, writing and parsing configuration), as well as your very first MCMC implementation! Congratulations :tada: 


I think my main advice would be to use more precisely typed objects than `dict`s and `**kwargs` (e.g., Pydantic's `BaseModel`), but using them can also be beneficial when we move some of the infrastructure to BlackJAX (JAX doesn't work well with Pydantic, but supports dictionaries.)

pawel-czyz (approved):
Very nice! I think there's quite a few new tricks (logging, xarray, writing and parsing configuration), as well as your very first MCMC implementation! Congratulations :tada: 


I think my main advice would be to use more precisely typed objects than `dict`s and `**kwargs` (e.g., Pydantic's `BaseModel`), but using them can also be beneficial when we move some of the infrastructure to BlackJAX (JAX doesn't work well with Pydantic, but supports dictionaries.)

workflows/config/mcmc_config_mark00.json
gordonkoehn: Yes, please teach me. Would like to learn the good way. Perhaps something for the next Sprint. Will make an issue, so we don't forget. See #35.

scripts/cell_simulation.py
gordonkoehn: Yes, you mentioned it before. I should replace the other usages of `os` with `pathlib` in the future.

scripts/run_mcmc.py
gordonkoehn: Thank you - this means a lot to me!

scripts/run_mcmc.py
gordonkoehn: Acknowledged. Thanks.

src/pyggdrasil/tree_inference/_mcmc_sampler.py
gordonkoehn: Changed to 4 digits precision.

src/pyggdrasil/tree_inference/_mcmc_sampler.py
gordonkoehn: Ok, hope we get to this at some stage. Interested to see how fast JAX is in relation.

src/pyggdrasil/tree_inference/_mcmc_sampler.py
gordonkoehn: Oh perfect. This solves my issue. I only used `**kwargs` to pass a single argument optionally without `pyright` complaining. - This is BETTER !

Issue #22, opened 2023-04-19
Add a function taking two TreeNodes and returning distance


pawel-czyz:
Time estimates: (S and M) or (M and M), depending how problematic the distance functions are.

PR #29, opened 2023-04-29
Add utilities from scPhylo

### LINKED ISSUE #2: Add HUNTRESS
[HUNTRESS](https://github.com/PASSIONLab/HUNTRESS) is a fast tree inference method. We want to have an implementation of it, so that it can be used from a specific module and is compatible with the rest of the data types used in this project.

Added to sprint #18

pawel-czyz:
This PR adds a dependency on (a fork of) the scPhylo library. In particular, we can now:

  - Use distance/similarity functions.
  - Use tree inference methods as HUNTRESS.

Hence, it resolves #2 and #22.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,8 +11,10 @@ python = "^3.9"
 anytree = "^2.8.0"
 jax = "^0.4.4"
 jaxlib = "^0.4.4"
-numpy = "^1.24.2"
+numpy = "^1.23.0"
 typing-extensions = "^4.5.0"
+llvmlite = "^0.39.1"
+scphylo-tools = {git = "https://github.com/pawel-czyz/scphylo-tools.git"}
 
 
 [tool.poetry.group.dev.dependencies]
--- a/tests/distances/test_scphylo_wrapper.py
+++ b/tests/distances/test_scphylo_wrapper.py
@@ -1,0 +1,32 @@
+"""Tests of distance and similarity functions
+imported from scphylo"""
+import numpy as np
+import pandas as pd
+import pytest
+import scphylo
+
+
+def test_basic() -> None:
+    """TEMPORARY: to be adjusted."""
+    genotypes = np.asarray(
+        [
+            [0, 0, 0, 0],
+            [1, 0, 0, 0],
+            [1, 0, 0, 1],
+            [1, 0, 1, 0],
+            [0, 1, 0, 0],
+        ]
+    )
+
+    df = pd.DataFrame(genotypes)
+
+    df.columns = [str(x) for x in df.columns]
+    df.index = [str(x) for x in df.index]
+
+    sim = scphylo.tl.mp3(df, df)
+
+    assert sim == pytest.approx(1.0)
+
+    rec = scphylo.tl.huntress(df, 0.001, 0.001)
+
+    assert isinstance(rec, pd.DataFrame)


src/pyggdrasil/distances/_interface.py
```diff
@@ -7,6 +7,37 @@
 _IntegerTreeRoot: TypeAlias = TreeNode[int, Any]
 
 
+class TreeSimilarity(Protocol):
```
gordonkoehn, line 10: Currently, we only use `TreeSimilarity` correct? -`TreeDistance` is here only for future tree dissimilarity measures to be implemented?

gordonkoehn (approved):
Looks good to me - Thank you for building this! - Added two minor comments only.
1) the format of the mutation matrix passed to the huntress inference, differed from our usual format I believe
2) a question/doubt about the default return values of the `TreeSimilarity` Protocols you defined

src/pyggdrasil/tree_inference/_huntress.py
```diff
@@ -0,0 +1,58 @@
...
+
+import anytree
+import numpy as np
+import pandas as pd
+import scphylo
+import pyggdrasil._scphylo_utils as utils
+
+
+def huntress_tree_inference(
+    mutations: np.ndarray,
```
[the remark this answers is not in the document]
pawel-czyz: Good points!

My understanding is that HUNTRESS:
  - doesn't handle homozygous entries,
  - I'm unsure if it can handle missing entries,
so for now we assume that the entries are 0 or 1. (As said in docstring).


Regarding the shape, we need to make sure we have one convention and we stick to it (also documenting it or annotating the shapes with `jaxtyping`). For example, here the docstring suggests that it's `(cells, sites)`, which indeed is the transpose of the convention you mentioned:  
```
def attach_cells_to_tree(
    rng: interface.JAXRandomKey,
    tree: interface.TreeAdjacencyMatrix,
    n_cells: int,
    strategy: CellAttachmentStrategy,
) -> PerfectMutationMatrix:
    """Attaches cells to the mutation tree.

    Args:
        rng: JAX random key
        tree: matrix representing mutation tree
            with the highest index representing the root
        n_cells: number of cells to sample
        strategy: cell attachment strategy.
          See ``CellAttachmentStrategy`` for more information.

    Returns:
        binary matrix of shape ``(n_cells, n_sites)``,
          where ``n_sites`` is determined from the ``tree``
          NOTE: Last row will be all ones is the root node.
          NOTE: not truncated the last row as shown in the SCITE paper
    """
    ...
```

tests/test_scphylo_utils.py
```diff
@@ -0,0 +1,65 @@
...
+    """
+    root = anytree.Node("root")
+    tp53 = anytree.Node("TP53", parent=root)
+    anytree.Node("KRAS", parent=root)
+    anytree.Node("NRAS", parent=tp53)
+    anytree.Node("BRCA", parent=tp53)
+    return root
+
+
+@pytest.fixture
```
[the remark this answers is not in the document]
pawel-czyz: Nice! :raised_hands:

src/pyggdrasil/distances/_interface.py
pawel-czyz: Yes – I adjusted `TreeDistance` protocol, so that all these methods have `calculate: (tree1, tree2) -> float` method. We will be able to use them interchangeably when drawing the plots.

(Later we may see that some of them are more suitable for diagnosing convergence issues, but right now it's too early to say which one to go with...)

src/pyggdrasil/distances/_interface.py
```diff
@@ -7,6 +7,37 @@
...
+
+    def is_symmetric(self) -> bool:
+        """Returns ``True`` if the similarity function is symmetric,
+        i.e., :math:`s(t_1, t_2) = s(t_2, t_1)` for all pairs of trees.
+
+        Note:
+            If it is not known whether the similarity function is symmetric,
+            ``False`` should be returned.
+        """
+        return True
```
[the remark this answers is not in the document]
pawel-czyz: Good point! Fixed :slightly_smiling_face:

src/pyggdrasil/_tree_utils.py
```diff
@@ -0,0 +1,21 @@
...
+
+    # Sort children by their tag to compare without order assumption
+    sorted_children1 = sorted(tree1.children, key=lambda x: x.name)
+    sorted_children2 = sorted(tree2.children, key=lambda x: x.name)
+
+    for child1, child2 in zip(sorted_children1, sorted_children2):
+        if not compare_trees(child1, child2):
+            return False
+
+    return True
```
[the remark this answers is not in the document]
pawel-czyz: Thanks! To be fair, the first draft of this method was created by GPT, I only fixed some bugs :wink:

src/pyggdrasil/tree_inference/_huntress.py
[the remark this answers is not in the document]
pawel-czyz: I opened #31 to address this problem. We can choose the convention at the meeting and include this in the next sprint.

src/pyggdrasil/distances/_scphylo_wrapper.py
```diff
@@ -0,0 +1,63 @@
+"""Thin wrappers around scPhylo distance and similarity metrics."""
+import scphylo
```
gordonkoehn, line 2: Do we have `scphylo` directly as a dependency? - I thought we only have your modified fork `scphylo-tools`. Tried to run it locally and it fails.

pyproject.toml
```diff
@@ -11,8 +11,10 @@ python = "^3.9"
 anytree = "^2.8.0"
 jax = "^0.4.4"
 jaxlib = "^0.4.4"
-numpy = "^1.24.2"
+numpy = "^1.23.0"
 typing-extensions = "^4.5.0"
+llvmlite = "^0.39.1"
+scphylo-tools = {git = "https://github.com/pawel-czyz/scphylo-tools.git"}
```
gordonkoehn, line 17: When trying to install this branch I get some issues with installing these two dependencies. Any idea where the problem may arise from?

[Screenshot 2023-05-04 at 10 57 17](https://user-images.githubusercontent.com/10867778/236158109-3fff70be-2d49-4ba3-9a00-d83d8b9f3207.png)
[Screenshot 2023-05-04 at 10 57 33](https://user-images.githubusercontent.com/10867778/236158114-29ab7d4a-d1fe-4a6c-82c9-d74402a89b3c.png)

src/pyggdrasil/visualize/tree.py
```diff
@@ -67,7 +67,7 @@ def plot(
         dot_string += line
     # convert to networkX graph
     graphs = pydot.graph_from_dot_data(dot_string)
-    graph = graphs[0]
+    graph = graphs[0]  # type: ignore
```
gordonkoehn, line 70: Took the liberty to suppress a false positive type warning here, that was causing your builts to fail.

src/pyggdrasil/distances/_scphylo_wrapper.py
pawel-czyz: `scphylo-tools` repository hosts Python package which is imported as `scphylo`. I guess this convention was taken from https://scvi-tools.org/

pyproject.toml
pawel-czyz: Hmmm... perhaps installing `llvmlite` manually could help? Or if Poetry can't handle the dependency, you can try to install (in the Poetry environment, so `poetry run pip install llvmlite` or via `poetry shell`) it using plain `pip`

src/pyggdrasil/visualize/tree.py
pawel-czyz: Wow, thanks a lot! :slightly_smiling_face:

pyproject.toml
gordonkoehn: managed to install `llvmlite` now in my poetry env now. - Not sure what worked in the end. 

I unfortunately I still can't run `poetry pytest` successfully, `scphylo`is still not recognized - despite my appempt of manually git clone and install it into poetry dirs directly.  Any suggestions?

pawel-czyz: Do you happen to have original `scphylo` installed (can be visible as `scphylo-tools`)? Perhaps it already sees that there's a package like this and doesn't take the version from git.

Alternatively, you can try something like `poetry run pip install "scphylo @ git+https://github.com/pawel-czyz/scphylo-tools.git"` or `poetry run pip install "scphylo-tools @ git+https://github.com/pawel-czyz/scphylo-tools.git"`)

Issue #30, opened 2023-05-01
Consistent naming of FNR and FPR


pawel-czyz:
In the original SCITE paper $\alpha$ stands for FPR (false positive rate) and $\beta$ stands for FNR (false negative rate).

However, in the cell simulation script the notation is reversed:

```python
    parser.add_argument("--alpha", required=True, help="False negative rate", type=float)
    parser.add_argument("--beta", required=True, help="False positive rate", type=float)
```
It'd be good to go through the code and see whether the naming is consistent. I'd also perhaps suggest to drop $\alpha$ and $\beta$ notation entirely and use `fpr` and `fnr` instead.

gordonkoehn:
Time Estimate: S

gordonkoehn:
The correct notation as in the SCITE paper is: 

alpha = $\alpha$ = False Positive Rate  = FPR
beta = $\beta$ = False Negative Rate = FNR

PR #36, opened 2023-05-12
Consistent naming of FNR and FPR

gordonkoehn:
> In the original SCITE paper stands for FPR (false positive rate) and stands for FNR (false negative rate).
>  
>  It'd be good to go through the code and see whether the naming is consistent. I'd also perhaps suggest dropping 
>  and notation entirely and use fpr and fnr instead.

see / resolves Issue #30

### DIFF
--- a/src/pyggdrasil/tree_inference/_simulate.py
+++ b/src/pyggdrasil/tree_inference/_simulate.py
@@ -2,14 +2,17 @@
 from enum import Enum
 
 import numpy as np
-from jax import random
 import jax.numpy as jnp
+import logging
+
+from jax import random
+from pydantic import BaseModel, validator
 
 import pyggdrasil.serialize as serialize
 import pyggdrasil.tree_inference._interface as interface
 from pyggdrasil.tree import TreeNode
 
-from typing import Union, TypedDict
+from typing import Union
 from jax import Array
 
 # Mutation matrix without noise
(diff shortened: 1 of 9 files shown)


src/pyggdrasil/tree_inference/_simulate.py
```diff
@@ -587,15 +587,27 @@ def adjacency_to_root_dfs(
...
-    params: dict,
+    params: CellSimulationParams,
     rng: interface.JAXRandomKey,
 ) -> dict:
     """
     Generates cell mutation matrix for one tree.
 
     Args:
-        params: input parameters from parser
+        params: TypedDict from parser of cell_simulation.py
```
gordonkoehn, line 610: Reading this code again - now with the knowledge about `TypedDict`s - I felt the urge to do at least a little bit of type checking/enforcing parameters here.

pawel-czyz (approved):
Wonderful! Please, merge :slightly_smiling_face:

src/pyggdrasil/tree_inference/_simulate.py
pawel-czyz: For validation I really recommend [Pydantic](https://docs.pydantic.dev/latest/usage/models/), it's really excellent.

gordonkoehn: Added a Pydantic model now :)

PR #40, opened 2023-05-17
Resolve dependency issue

pawel-czyz:
The problem was with `cyvcf2` dependency, which yesterday was updated from 0.3.19 to 0.3.20, causing our build to fail. Hence, we temporary pin it to 0.3.19, hoping that the issue will be resolved over the next few weeks.


gordonkoehn:
Woah -  Fantastic! - Thank you for the speedy resolution !! - I recall seeing `cyvcf2`appearing in the errors a bunch, will look at error messages more closely next time.

PR #42, opened 2023-05-24
Visualisation of MCMC iteration/time series 

gordonkoehn:
Building upon #38, here we are building the actual plots. 

The envisioned product. 

- [x] 1) The plot with the X-axis being the iteration number, and Y axis being the loglikelihood of the sampled tree.
- [x] 2) The number of iterations needed to sample the true tree. - to some JSON/ txt
- [x] 3) Plot of the distance to the true tree on the Y axis for several chains and the iteration number on the X axis.

This PR is the 2/2 for closing issue #23.

PR #49, opened 2023-05-30
Healing the tree of life - fixing _swap_subtrees_move_ for same lineage moves. 

### LINKED ISSUE #47: Consider cleaning up `tree_utils`
In `tree_inference.tree_utils` there are 3 functions from building the tree moves that are not used. Only used for testing purposes. 

`_prune`, `_reattach`, `prune_and_reattach_move`

One could consider moving these to tests, to keep the main code short.

### LINKED ISSUE #48: Investigate Potential Bug in MCMC Tree Moves
During the visualization issue #42 some trees in an outdated, hence now irreproducible dataset, outputted by the MCMC sampler were not convertible to the `TreeNode` format. The reason was, that they seemed to lack a root. - Or the last node in the adjacency matrix at least was not the root. The exact issue is not clear. The function `tree_inference_tree._get_root_label` got an empty array in their line 
`# find row which has all ones in ancestor_matrix
    root_idx = jnp.where(jnp.all(ancestor_matrix == 1, axis=1))[0]`
    
- [x] Consider adding debug logging to the tree moved to discover what went wrong the next time this happens.
- [x] Add tests for the individual tree moves and check if it is a valid tree
- [x] identify the issue
- [x] fix the issue
- [x] rerun mcmc mark00
- [x] check if all trees can be viszualized
- [x] check speed of mcmc

Conjectures:
- root is assumed to be the last in the tree, and was swapped to be not the last
- confusion if to include the root in moves or not

gordonkoehn:
This PR addresses/closes #48 mainly, it also closes #47. 

The `tree_inference.mcmc._swap_subtrees_move` and `tree_inference.mcmc._swap_subtrees_proposal` were incomplete/wrong for the case of choosing nodes of the same lineage. Now fixed and tested against the SCITE supplement paper examples. 

The failure of this tree move became apparent during the coding of the _Visualization of the Timeseries_ in #42, as some trees were broken into two and mcmc samples could not be imported.

To convince myself the tree moves now make sense and the trees don't break, I decided to include a small visualization script of the trees of a mcmc chain in this PR. This resulted in a very satisfying animation. See here: 

animation

These are the main objectives of this PR. 

It took a bit of tracing of the issues, hence note the addition of `loggers` in many scripts. 

Merging this PR will allow us to proceed with the  _Visualization of the Timeseries_ in #42.

Apologies for the large and messy PR - the result of some digging for the bug and the self-convincing of its workings.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,6 +22,7 @@ datetime = "^5.1"
 pytz = "^2023.3"
 pydantic = "^1.10.7"
 cyvcf2 = "<0.30.20"  # Temporary dependency: see https://github.com/cbg-ethz/PYggdrasil/pull/40/
+tqdm = "^4.65.0"
 
 
 [tool.poetry.group.dev.dependencies]
--- a/scripts/plot_trees.py
+++ b/scripts/plot_trees.py
@@ -28,7 +28,7 @@ import logging
 
 
 from pathlib import Path
-
+from tqdm import tqdm
 
 import pyggdrasil.serialize as serialize
 import pyggdrasil.tree_inference as tree_inf
@@ -97,14 +97,10 @@ def main() -> None:
     fullpath_d = Path(args.data_fp)
     # plot the trees
     mcmc_samples = serialize.read_mcmc_samples(fullpath=fullpath_d)
-    print(mcmc_samples)
     pure_data = tree_inf.to_pure_mcmc_data(mcmc_samples)
 
     # get iterations
     iterations = pure_data.iterations
-    print(iterations)
-    print(len(iterations))
-    print(len(pure_data.trees))
 
     # print options
     print_options = dict()
@@ -112,34 +108,29 @@ def main() -> None:
     print_options["data_tree"] = dict()
     print_options["data_tree"]["log-likelihood"] = True
 
-    j = 0
     # for each iteration, plot the tree
-    for i in iterations:
-        i = int(i)
-        j += 1
-        print("j = ", j)
-        print(i)
+    for i in tqdm(iterations):
         # get the sample
         sample = pure_data.get_sample(i)
-        print(sample)
         # get the tree
         tree = sample[1]
-        tree.print_topo()
         tree.data = dict()
         tree.data["log-likelihood"] = sample[2]
         # get the log probability, and round to 2 decimal places
         log_prob = sample[2]
         log_prob = round(log_prob, 2)
-        # make savename from iteration number
+        # make save name from iteration number
 
-        save_name = "iteration_" + str(i) + "_log_prob_" + str(log_prob) + ".svg"
+        save_name = "iter_" + str(i) + "_log_prob_" + str(log_prob)
 
         # make full path with pathlib
         save_dir = Path(args.out_dir)
 
         # plot tree
         visualize.plot(tree, save_name, save_dir, print_options)
-
+        # log
+        logging.info("Plotted tree for iteration %s", i)
+    # log end
     logging.info("End Session")
 
 
--- a/scripts/plot_trees.py
+++ b/scripts/plot_trees.py
@@ -101,7 +101,6 @@ def main() -> None:
     iterations = pure_data.iterations
     # convert iterations to list of integers
     iterations = list(map(int, iterations))
-
     # print options
     print_options = dict()
     print_options["title"] = False
--- a/scripts/plot_trees.py
+++ b/scripts/plot_trees.py
@@ -7,19 +7,9 @@ Example Usage:
 poetry run python ../scripts/plot_trees.py
 
 Example Usage with arguments:
-    poetry run python scripts/run_mcmc.py
-    --config_fp data/config/mcmc_config_mark00.json
-    --out_dir data/mcmc/mark01
-    --data_fp data/mock/seed_32_n_..._tree_3.json
-
-    if you want to provide a tree:
-    from a mcmc run:
-    --init_tree_fp data/mcmc/mark01/samples_XXXXXXXX_XXXXXX.json
-    --iteration 1000
-    or to read in a TreeNode
-    --init_tree_fp data/trees/tree_3.json
-    --init_TreeNode
-
+    poetry run python scripts/plot_trees.py
+    --out_dir data/plots/tree/mark00
+    --data_fp data/mcmc/mark00/samples.json
 
 """
 
--- a/scripts/plot_trees.py
+++ b/scripts/plot_trees.py
@@ -75,11 +75,17 @@ def main() -> None:
     # if logfile does not exist, create it
     if not Path(fullpath_log).exists():
         Path(fullpath_log).touch()
-    logging.basicConfig(filename=fullpath_log, level=logging.DEBUG)
+    # if logfile exist delete it and create a new one
+    else:
+        Path(fullpath_log).unlink()
+        Path(fullpath_log).touch()
+
+    logging.basicConfig(filename=fullpath_log, level=logging.INFO)
     # set logging level for jax
     logging.getLogger("jax._src.dispatch").setLevel(logging.ERROR)
     logging.getLogger("jax._src.interpreters.pxla").setLevel(logging.ERROR)
     logging.getLogger("jax._src.xla_bridge").setLevel(logging.ERROR)
+    logging.getLogger("matplotlib.texmanager").setLevel(logging.ERROR)
 
     logging.info("Starting Session")
 
--- a/scripts/plot_trees.py
+++ b/scripts/plot_trees.py
@@ -74,18 +74,13 @@ def main() -> None:
     # get the full path
     out_dir = Path(args.out_dir)
     # if out_dir does not exist, create it
-    if not out_dir.exists():
-        out_dir.mkdir(parents=True, exist_ok=True)
+    out_dir.mkdir(parents=True, exist_ok=True)
 
     # Set up logging
     fullpath_log = out_dir / "plot_trees.log"
     # if logfile does not exist, create it
-    if not Path(fullpath_log).exists():
-        Path(fullpath_log).touch()
-    # if logfile exist delete it and create a new one
-    else:
-        Path(fullpath_log).unlink()
-        Path(fullpath_log).touch()
+    Path(fullpath_log).unlink(missing_ok=True)
+    Path(fullpath_log).touch()
 
     logging.basicConfig(filename=fullpath_log, level=logging.INFO)
     # set logging level for jax
--- a/scripts/plot_trees.py
+++ b/scripts/plot_trees.py
@@ -103,17 +103,22 @@ def main() -> None:
     # get iterations
     iterations = pure_data.iterations
     print(iterations)
+    print(len(iterations))
+    print(len(pure_data.trees))
 
     # print options
     print_options = dict()
-    print_options["title"] = True
+    print_options["title"] = False
     print_options["data_tree"] = dict()
     print_options["data_tree"]["log-likelihood"] = True
-    print_options["data_tree"]["Data type"] = False
-    print_options["data_tree"]["Run"] = False
 
+    j = 0
     # for each iteration, plot the tree
     for i in iterations:
+        i = int(i)
+        j += 1
+        print("j = ", j)
+        print(i)
         # get the sample
         sample = pure_data.get_sample(i)
         print(sample)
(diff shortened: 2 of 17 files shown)


src/pyggdrasil/visualize/tree.py
```diff
@@ -71,8 +81,22 @@ def plot(
...
+    node_height = 0.75  # Height of each node in the figure
+    figure_width = width * node_width
+    figure_height = depth * node_height
+
+    # print(f"figure_width: {figure_width}")
+    # print(f"figure_height: {figure_height}")
+
     # plot
-    fig = plt.figure()
+    fig = plt.figure(figsize=(figure_width, figure_height))
```
gordonkoehn, line 99: Dynamic figure sizing depending on tree depth and breadth! Rather proud of this chatGPT-assisted gimmick. Works well with our 50-node trees.

tests/tree_inference/test_mcmc.py
```diff
@@ -191,3 +193,161 @@ def test_swap_subtrees_move():
...
+
+    # check if any of the four possible trees is the same as the new tree
+    assert (
+        jnp.array_equal(new_tree.tree_topology, new_tree_corr1.tree_topology)
+        or jnp.array_equal(new_tree.tree_topology, new_tree_corr2.tree_topology)
+        or jnp.array_equal(new_tree.tree_topology, new_tree_corr3.tree_topology)
+        or jnp.array_equal(new_tree.tree_topology, new_tree_corr4.tree_topology)
+    )
+    # check if the labels are the same, just once for sanity
+    assert jnp.array_equal(new_tree.labels, new_tree_corr1.labels)
```
gordonkoehn, line 353: Implements the different lineage subtree swap as in the SCITE supplement. 

Screenshot 2023-05-30 at 13 02 38

tests/visualize/test_viz_tree.py
```diff
@@ -9,7 +9,7 @@
 # from pathlib import Path
 
 from pyggdrasil.tree_inference._tree import Tree
-import pyggdrasil.visualize.tree as viz
+import pyggdrasil.visualize as viz
```
gordonkoehn, line 12: fixing public API

gordonkoehn (commented):
Ready for review - refactorings may distract from the main changed. Main changes are marked.

pyproject.toml
```diff
@@ -22,6 +22,7 @@ datetime = "^5.1"
 pytz = "^2023.3"
 pydantic = "^1.10.7"
 cyvcf2 = "<0.30.20"  # Temporary dependency: see https://github.com/cbg-ethz/PYggdrasil/pull/40/
+tqdm = "^4.65.0"
```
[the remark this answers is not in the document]
pawel-czyz: Yes, also tqdm has a `disable` argument, which can hide it.

scripts/plot_trees.py
```diff
@@ -0,0 +1,135 @@
...
+    Returns:
+        None
+    """
+
+    args = create_parser()
+
+    # get the full path
+    out_dir = Path(args.out_dir)
+    # if out_dir does not exist, create it
+    if not out_dir.exists():
```
pawel-czyz, line 70: I don't really get why the `if` statement is needed here – wouldn't `exist_ok=False` work here?

scripts/plot_trees.py
```diff
@@ -0,0 +1,135 @@
...
+        out_dir.mkdir(parents=True, exist_ok=True)
+
+    # Set up logging
+    fullpath_log = out_dir / "plot_trees.log"
```
pawel-czyz, line 74: If `out_dir` is a `Path`, then `fullpath_log` is also a `Path` and manual casting below is not really needed :wink:

scripts/plot_trees.py
```diff
@@ -0,0 +1,135 @@
...
+    if not Path(fullpath_log).exists():
+        Path(fullpath_log).touch()
+    # if logfile exist delete it and create a new one
+    else:
+        Path(fullpath_log).unlink()
+        Path(fullpath_log).touch()
+
+    logging.basicConfig(filename=fullpath_log, level=logging.INFO)
+    # set logging level for jax
+    logging.getLogger("jax._src.dispatch").setLevel(logging.ERROR)
```
pawel-czyz, line 85: I see what this is for, nice!

scripts/plot_trees.py
```diff
@@ -0,0 +1,135 @@
...
+
+    # print options
+    print_options = dict()
+    print_options["title"] = False
+    print_options["data_tree"] = dict()
+    print_options["data_tree"]["log-likelihood"] = True
+
+    # for each iteration, plot the tree
+    for i in tqdm(iterations):
+        i = int(i)
```
pawel-czyz, line 109: I don't think this `int(i)` is really needed, is it? `tqdm` should return just the original object.

src/pyggdrasil/interface/__init__.py
```diff
@@ -47,10 +56,13 @@ def get_sample(self, iteration: int) -> tuple[int, TreeNode, float]:
...
+        # get index of iteration
+        iteration_idx = jnp.where(self.iterations == iteration)[0][0]
+
         return (
             iteration,
-            self.trees[iteration],
-            self.log_probabilities[iteration].item(),
+            self.trees[iteration_idx],
+            self.log_probabilities[iteration_idx].item(),
         )
```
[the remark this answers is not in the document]
pawel-czyz: This is subtle and I'd really like to understand this: why `iteration_idx` can be different from `iteration`? Is it off by one error?

```
iteration_idx = jnp.where(self.iterations == iteration)[0][0]
```

src/pyggdrasil/tree_inference/_analyze.py
```diff
@@ -2,13 +2,18 @@
...
 from pyggdrasil import TreeNode, compare_trees
 from pyggdrasil.tree_inference import Tree, unpack_sample
 
 from pyggdrasil.interface import MCMCSample, PureMcmcData
 
 
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.DEBUG)
+
+
```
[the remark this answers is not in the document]
pawel-czyz: Very good idea :slightly_smiling_face: Indeed, if each module has a separate logger it's easier to see what the program is doing.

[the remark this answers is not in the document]
pawel-czyz: By the way, you may like reading about https://www.structlog.org/en/stable/ (even if it doesn't get used here – I used it once, when I was working on a large pipeline and automated analysis of any failures/caught exceptions/whatever unexpected happened was essential).

src/pyggdrasil/tree_inference/_mcmc.py
```diff
@@ -39,6 +42,11 @@ def _prune_and_reattach_move(tree: Tree, *, pruned_node: int, attach_to: int) ->
     # make new tree
     new_tree = Tree(tree_topology=new_adj_mat, labels=tree.labels)
 
+    logger.debug(
+        "MCMC: Prune and reattach move - pruned node %s; attached to node %s",
+        pruned_node,
+        attach_to,
+    )
```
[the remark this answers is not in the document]
pawel-czyz: I don't think it impacts the speed by any significant factor.
I think it's wonderful that you added them and traced the bug, so unless we need to optimise we can leave them here. (And if we want to optimise, we'll go to the fully JAX-compliant implementation anyway, I think.)

src/pyggdrasil/tree_inference/_mcmc.py
```diff
@@ -124,15 +138,23 @@ def _swap_node_labels_proposal(
...
-def _swap_subtrees_move(tree: Tree, node1: int, node2: int) -> Tree:
-    """Swaps subtrees rooted at ``node1`` and ``node2``.
+def _swap_subtrees_move(
+    tree: Tree,
+    node1: int,
+    node2: int,
+    same_lineage: bool,
+    key: random.PRNGKeyArray = None,
+) -> Tree:
+    """Swaps subtrees rooted at ``node1``/ ``node i`` and ``node2``/ ``node k``.
```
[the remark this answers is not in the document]
pawel-czyz: This answer is going to be very subjective and opinionated, but:
  - in Python overloading is very limited, although sometimes useful for type checking – see e.g., [this tutorial](https://adamj.eu/tech/2021/05/29/python-type-hints-how-to-use-overload/);
  - whenever different sets of arguments are passed, it is a potential sign that *this function does two things and should be refactored into two functions*. (Recall [SRP](https://en.wikipedia.org/wiki/Single-responsibility_principle)).

I think this is a good example: if the case of being in the same lineage is so important, it should perfectly be in a different function. It will simplify the logic.

Then, the move (which has access to the random key) can decide which of these functions should be used (by comparing the nodes and observing whether they are in the same lineage).

I'm not sure if this writing is clear and coherent enough – let me know if not, we can also discuss it in the meeting :slightly_smiling_face:

src/pyggdrasil/tree_inference/_mcmc.py
```diff
@@ -145,14 +167,48 @@ def _swap_subtrees_move(tree: Tree, node1: int, node2: int) -> Tree:
...
+        # make new tree
+        new_tree = Tree(tree_topology=new_adj_mat, labels=tree.labels)
+
+        logger.debug(
+            "MCMC: Swap subtrees move - nested swap of"
+            " subtrees rooted at node %s and node %s",
+            node1,
+            node2,
+        )
+        return new_tree
```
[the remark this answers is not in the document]
pawel-czyz: Great you've found it!

I should've actually caught that in the review implementing the move. I need to improve on my false negative rate :sweat_smile:

src/pyggdrasil/visualize/tree.py
```diff
@@ -55,6 +59,12 @@ def plot(
     Note:
         see tests/visualize/test_tree.py for example usage
     """
+
+    # check if tree-name is none and throw log - and error
+    if tree.name is None and print_options["title"] is True:
+        logger.error("Tree name is None, cannot print title")
+        raise ValueError("Tree name is None, cannot print title")
+
```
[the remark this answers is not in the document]
pawel-czyz: I see :slightly_smiling_face:

I'm not sure if it'd be cleaner in this case, but sometimes using a non-union type may be also an option (e.g., `name` could always be a `str` or set by default to `""`).

src/pyggdrasil/visualize/tree.py
pawel-czyz: Oh, this is very cool! Now I see how you generated this video! :slightly_smiling_face: 


@monicadragan I know that you have a whole JS library for drawing trees, but perhaps it'd be of interest also to you :slightly_smiling_face:

tests/tree_inference/test_mcmc.py
pawel-czyz: Great job! :slightly_smiling_face:

pawel-czyz (approved):
This is *really good*. I think it's great you found this bug and as a side effect you added logging and visualisation utilities.

My only real comment (and which doesn't need to be addressed before merging) is that the swap subtree move should really be based on two functions: one for nodes in the same lineage and the other one. They take different arguments and have different logic, what can be seen in the long `if`/`else` block.

But the current version works and you added a nice test from the paper – looks as ready to merge! Great work :slightly_smiling_face:

src/pyggdrasil/tree_inference/_mcmc.py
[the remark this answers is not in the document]
gordonkoehn: Great, would be excited to see how the optimization with Jax would work. :)

src/pyggdrasil/visualize/tree.py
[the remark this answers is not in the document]
gordonkoehn: Yes, I see. - I think I'll keep it this way, just to be explicit.

src/pyggdrasil/visualize/tree.py
gordonkoehn: Still experimental - not yet tested on all sensible tree sizes.

scripts/plot_trees.py
gordonkoehn: Good point! Made a few changes to be more efficient here.

scripts/plot_trees.py
gordonkoehn: Oh great. - overlooked this.

scripts/plot_trees.py
gordonkoehn: Only really needed if one sets the root logging to `DEBUG`. But I left it, for convenience for future bugs.

src/pyggdrasil/tree_inference/_mcmc.py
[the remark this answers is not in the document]
gordonkoehn: Very interesting! As always, I greatly apparate these nuggets of new input :) 

Tried to comply with SRP now. Split the function up into 3. 
1) `detrach the subtrees` 2) same lineage 3) different lineage

But still need to keep the choice of same or different lineage in the `swap_subtrees_proposal`, needed for the correction factor.

gordonkoehn (commented):
all good now. -  all feedback in.

Issue #54, opened 2023-06-06
Unique, Flexible File naming for Trees, Cell Simulations and MCMC Samples


gordonkoehn:
Devise a consistent file naming for generated trees, cell_simulation,mcmc_run that is short enough to be filenames, but unique and flexible to allow to all thinkable snakemake workflows

Issue #53, opened 2023-06-05
Devise a scalable workflow / structure to analyze and save all


gordonkoehn:
To devise a scalable, modular, reproducible and extendable workflow design that encompasses:

- tree generation
- cell simulation
- MCMC run
- analysis for true tree and distances for single and multiple runs
- visualization

gordonkoehn:
- [ ] devise a consistent file naming for `generated trees`, `cell_simulation`,`mcmc_run` that is short enough to be filenames, but unique and flexible to allow to all thinkable snakemake workflows

gordonkoehn:
-[ ]  devise a consistent file naming for generated trees, cell_simulation,mcmc_run that is short enough to be filenames, but unique and flexible to allow to all thinkable snakemake workflows

PR #55, opened 2023-06-06
Creating modular ID classes for (Trees, Cell Simulations, MCMC Run) for file naming

gordonkoehn:
This PR addresses #54 and is part of designing #53.

For a scalable snakemake workflow, I understand I need unique IDs for all my files. The current filename, whilst human readable and unique - with some flaws - results in very long filenames and is not standardized between classes - worse the naming is done is the scripts. This PR aims to create unified capabilities for all scripts ( perhaps even the `serialize` module) to call to create IDs to be incorporated into filenames. - so that snakemake has it easy. 

IDs are composed of concatenating all parameters of generation including seeds - making it reproducible - and with effort human readable - given one knows the order of parameters.

Derivatives of files needed to generate the former are carrying the generation file ID in their name. I.e. an MCMC sample will carry both an ID of the initial tree as well as the cell simulation data within it. 

This PR does not integrate the classes created here into the current workflow - will be done in future issues.

### Generated deep tree 
`T` indicates it's a tree, `d` the type of tree i.e. deep, of 10 nodes and a seed of 123
`T_d_10_123"`

### Cell simulation i.e. a mutation matrix
 generated given the above tree and a seed 42, 1000 cells, errors(0.01_0.02), a rate 0.03, homozygous mutations = true, cell attachment strategy UIR - uniform including the root 
`CS_42-T_d_10_123-1000_0.01_0.02_0.03_t_UIR`

Note how the tree is provided between the `-` carrying the information about the number of mutations =  nodes - 1

(If we ever switch to real data without a known true tree, the parent class `MutationDataId`of `CellSimulationId` may handle that and add the mutations argument back in without a tree.)

### MCMC samples
Given the tree and cell simulation data as above the mcmc samples and log will be named as follows. 

Further parameters: 
-  mcmc seed 42,
-  the initial tree here `iT_d_10_123` in the same format as above 
- Mcmc run config `MC_1.24e-06_0.097_1000_0_1` with (fpr: 1.24e-06, fnr: 0.097, n_samples: 1000, burn_in = 0, thinning = 1)
- Move Probailities Config `MPC_0.1_0.65_0.25` where the numbers are (prune_and_reattach, swap_node_labels,swap_subtrees)

The last two are implemented as `pydantic` models with default values and rudimentary checks.

The full mcmc run files have IDs like:

`MCMC_42-CS_42-T_d_10_123-1000_0.01_0.02_0.03_t_UIR-iT_d_10_123-MC_1.24e-06_0.097_1000_0_1-MPC_0.1_0.65_0.25`

### DIFF
--- a/src/pyggdrasil/tree_inference/_config.py
+++ b/src/pyggdrasil/tree_inference/_config.py
@@ -1,0 +1,63 @@
+"""Config classes for MCMC sampler,
+ cell simulation and tree inference, distance calculation."""
+
+from pydantic import BaseModel, validator
+
+
+class MoveProbConfig(BaseModel):
+    """Move probabilities for MCMC sampler."""
+
+    prune_and_reattach: float = 0.1
+    swap_node_labels: float = 0.65
+    swap_subtrees: float = 0.25
+
+    @validator("prune_and_reattach", "swap_node_labels", "swap_subtrees")
+    def move_prob_validator(cls, v):
+        """Probabilities sum to 1."""
+        total = sum(v.values())
+        if total != 1:
+            raise ValueError("Move probabilities must sum to 1")
+        return v
+
+    def id(self) -> str:
+        """String representation of move probabilities."""
+        str_rep = "MPC_" + str(self.prune_and_reattach)
+        str_rep = str_rep + "_" + str(self.swap_node_labels)
+        str_rep = str_rep + "_" + str(self.swap_subtrees)
+        return str_rep
+
+
+class McmcConfig(BaseModel):
+    """Config for MCMC sampler."""
+
+    move_probs: MoveProbConfig
+    fpr: float = 1.24e-06
+    fnr: float = 0.097
+    n_samples: int
+    burn_in: int = 0
+    thinning: int = 1
+
+    @validator("fpr")
+    def fpr_validator(cls, v):
+        """Validate move probabilities."""
+        if v <= 0 or v > 1:
+            raise ValueError("fpr must be between 0 and 1")
+        return v
+
+    @validator("fnr")
+    def fnr_validator(cls, v):
+        """Validate move probabilities."""
+        if v <= 0 or v > 1:
+            raise ValueError("fnr must be between 0 and 1")
+        return v
+
+    def id(self) -> str:
+        """String representation of MCMC config."""
+        str_rep = "MC"
+        str_rep = str_rep + "_" + str(self.fpr)
+        str_rep = str_rep + "_" + str(self.fnr)
+        str_rep = str_rep + "_" + str(self.n_samples)
+        str_rep = str_rep + "_" + str(self.burn_in)
+        str_rep = str_rep + "_" + str(self.thinning)
+        str_rep = str_rep + "_" + str(self.move_probs.id())
+        return str_rep
--- a/src/pyggdrasil/tree_inference/_config.py
+++ b/src/pyggdrasil/tree_inference/_config.py
@@ -1,21 +1,22 @@
 """Config classes for MCMC sampler,
  cell simulation and tree inference, distance calculation."""
 
-from pydantic import BaseModel, validator
+from pydantic import BaseModel, validator, confloat, conint
 
 
 class MoveProbConfig(BaseModel):
     """Move probabilities for MCMC sampler."""
 
-    prune_and_reattach: float = 0.1
-    swap_node_labels: float = 0.65
-    swap_subtrees: float = 0.25
+    prune_and_reattach: confloat(gt=0, lt=1) = 0.1  # type: ignore
+    swap_node_labels: confloat(gt=0, lt=1) = 0.65  # type: ignore
+    swap_subtrees: confloat(gt=0, lt=1) = 0.25  # type: ignore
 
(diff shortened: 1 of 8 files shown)


gordonkoehn:
@pawel-czyz I went bold to create a plethora of new classes here instead of functions - What do you think of this? Is this overcomplicating it or is there a much simpler way to handle so many dimensions of parameters? Is my file-naming approached even sensible?

src/pyggdrasil/tree_inference/_config.py
```diff
@@ -0,0 +1,63 @@
+"""Config classes for MCMC sampler,
+ cell simulation and tree inference, distance calculation."""
+
+from pydantic import BaseModel, validator
+
+
+class MoveProbConfig(BaseModel):
+    """Move probabilities for MCMC sampler."""
+
+    prune_and_reattach: float = 0.1
```
pawel-czyz, line 10: Glad you started using Pydantic! :slightly_smiling_face: 

It's not necessary to correct this, but in the future you may prefer using constrained floats (see [this](https://docs.pydantic.dev/latest/usage/types/#constrained-types))

src/pyggdrasil/tree_inference/_config.py
```diff
@@ -0,0 +1,63 @@
...
+    swap_node_labels: float = 0.65
+    swap_subtrees: float = 0.25
+
+    @validator("prune_and_reattach", "swap_node_labels", "swap_subtrees")
+    def move_prob_validator(cls, v):
```
pawel-czyz, line 15: Nice!

src/pyggdrasil/tree_inference/_config.py
```diff
@@ -0,0 +1,63 @@
...
+    """Config for MCMC sampler."""
+
+    move_probs: MoveProbConfig
+    fpr: float = 1.24e-06
+    fnr: float = 0.097
+    n_samples: int
+    burn_in: int = 0
+    thinning: int = 1
+
+    @validator("fpr")
```
pawel-czyz, line 40: Nice validator! (I also think you may appreciate `confloat` by looking how to implement it by adding ~10 character, rather than a few lines :wink: )

src/pyggdrasil/tree_inference/_config.py
```diff
@@ -0,0 +1,63 @@
...
+    @validator("fnr")
+    def fnr_validator(cls, v):
+        """Validate move probabilities."""
+        if v <= 0 or v > 1:
+            raise ValueError("fnr must be between 0 and 1")
+        return v
+
+    def id(self) -> str:
+        """String representation of MCMC config."""
+        str_rep = "MC"
```
pawel-czyz, line 56: This works! What I also like is using format strings, e.g.,
```
str_rep = f"MC_{self.fpr:.3f}_{self.fnr:.3f}_{self.n_samples}_etc"
```

src/pyggdrasil/tree_inference/_file_id.py
```diff
@@ -0,0 +1,206 @@
...
+Cell Simulation and MCMC run files uniquely """
+
+from enum import Enum
+from typing import Union, Optional
+
+
+from pyggdrasil.tree_inference import CellAttachmentStrategy, McmcConfig
+
+
+class TreeType(Enum):
```
pawel-czyz, line 11: Very good idea!

src/pyggdrasil/tree_inference/_file_id.py
```diff
@@ -0,0 +1,206 @@
...
+        """
+
+        self.tree_type = tree_type
+        self.n_nodes = n_nodes
+        self.seed = seed
+        self.cell_simulation_id = cell_simulation_id
+
+        self.id = self._create_id()
+
+    def _create_id(self) -> str:
```
pawel-czyz, line 78: It may stay this way, but it'd be cleaner not to have side effects in this method (as above you have `self.id = self._create_id()`, it could only return a value).

pawel-czyz (approved):
Very good PR! It's both clear to read and contains several cool programming tricks (as enums and Pydantic with its validators) :slightly_smiling_face:

src/pyggdrasil/tree_inference/_config.py
gordonkoehn: Great to know - thank you!

src/pyggdrasil/tree_inference/_config.py
gordonkoehn: Oh, yes -  implemented with the above `confloat` and precision.

src/pyggdrasil/tree_inference/_config.py
gordonkoehn: @pawel-czyz implemented it now with `confloat` but had to silence `pyright`

I run into issues with `pyright` - it does not recognize `confloat`. Do you know a workaround?

`Illegal type annotation: call expression not allowed (reportGeneralTypeIssues)`

No solution here : https://github.com/pydantic/pydantic/issues/5006

src/pyggdrasil/tree_inference/_file_id.py
gordonkoehn: Yes, I see -  changed it.

Issue #45, opened 2023-05-25
Temporary fixing of pyright version 


gordonkoehn:
It appears that out code has some `pyright` issues with the latest update. We are running into a bunch of issues, when trying to iterate through our tree objects: 
![Screenshot 2023-05-25 at 10 46 54](https://github.com/cbg-ethz/PYggdrasil/assets/10867778/7429a006-f6df-4bb3-aaed-cc8a29c381db)

Fortunately, we are not the only ones, there are two somewhat related issues from a quick look  - see https://github.com/microsoft/pyright/issues/5165 and https://github.com/microsoft/pyright/issues/5172

I fixed `pyright` to the latest working version - which for us is <1.1.309, i.e 1.1.3.08.

PR #58, opened 2023-06-07
Revert Temporary Fixed Version of Pyright and cyvcf2

gordonkoehn:
I was notified that the new pyright versions should no longer have that pase positive error we go a while back #45.

Already being on it, I decided to also check for the cyvcf2 version that we had also fixed and was causing problems #40


pawel-czyz (approved):
Thanks for taking care of that!

Issue #59, opened 2023-06-07
Refactor generate Trees to use TreeNode format as public API


gordonkoehn:
To shelter more and more complexity - the `Tree` or even worse the bare adjacency format should not leave the `tree_inference` module - hence the `tree_inference.tree_generator` should be rewritten to output the Trees generated as `TreeNode` in public API or `Trees` in the private API

PR #62, opened 2023-06-07
Refactor generate Trees to use TreeNode format as public API

gordonkoehn:
This PR addresses #59.

Some minor refactoring was done in `run_mcmc.py` just because `pyright` did just catch it here.

### DIFF
--- a/scripts/run_mcmc.py
+++ b/scripts/run_mcmc.py
@@ -189,7 +189,7 @@ def run_chain(
         rng_tree, rng = random.split(rng, 2)
         #  generate random trees (uniform sampling) as adjacency matrix
         #  / add +1 for root
-        tree = tree_inf.generate_random_tree(rng_tree, n_nodes=n_mutations + 1)
+        tree = tree_inf._generate_random_tree_adj_mat(rng_tree, n_nodes=n_mutations + 1)
         tree = jnp.array(tree)
         # make Tree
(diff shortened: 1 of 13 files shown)


scripts/run_mcmc.py
```diff
@@ -223,21 +223,21 @@ def run_chain(
...
-        error_rates=(config["fpr"], config["fnr"]),
+        error_rates=(config.fpr, config.fnr),
         move_probs=move_probs,
-        num_samples=config["num_samples"],
-        num_burn_in=config["burn_in"],
+        num_samples=config.n_samples,
+        num_burn_in=config.burn_in,
         output_dir=Path(params.out_dir),
-        thinning=config["thinning"],
+        thinning=config.thinning,
```
gordonkoehn, line 240: Minor refactoring not planned for this PR - yet pyright did only catch it here.

pawel-czyz: I see! I think it's a good change – stronger types are usually help with code readability and spotting hidden bugs :slightly_smiling_face:

pawel-czyz (approved):
Looks good to me! Just a question: are these functions backends to the functions covered by the unit tests you merged yesterday? If not, it'd be good to check whether they produce the right trees. (This is definitely one of these places where I would introduce a bug if I didn't write unit tests.)

Issue #61, opened 2023-06-07
Write Script for Tree Generation


gordonkoehn:
For modularity, the generation of Trees should be separate from the cell simulation. Hence, a script shall be written to allow for the generation of "random", "deep" and "star" trees given, if applicable, a seed and the number of nodes.

PR #63, opened 2023-06-07
Write Script for Tree Generation

gordonkoehn:
This PR addresses #61.

Used the newly implemented naming infrastructure from:
- #54 

Contains some placeholder functions, implemented in
- #62 

Should be final to be integrated into the final snakemake workflow.


pawel-czyz (approved):
Very nice! I don't have any suggestions, I wouldn't do it better myself :wink:

PR #64, opened 2023-06-07
Revised version of Cell Simulation

gordonkoehn:
In this PR, I revised the `cell_simulation.py` scripts for the mock data generation to 
- use the new file naming provided by `tree_inference.CellSimulationId` 
- generate one mock data set for one tree given (allowing to supply the true tree instead of adhock random generation)

### DIFF
--- a/scripts/cell_simulation.py
+++ b/scripts/cell_simulation.py
@@ -10,7 +10,7 @@ As per definition in the SCITE Jahn et al. 2016.
 Example Usage:
 poetry run python ../scripts/cell_simulation.py
 --seed 42
---out_dir ../data --n_trees 3 --n_cells 100 --n_mutations 8
+--out_dir ../data --n_cells 100 --init_tree ../data/T_d_10_123.json
  --strategy UNIFORM_INCLUDE_ROOT --fpr 0.01 --fnr 0.02
  --na_rate 0.01 --observe_homozygous True --verbose
 """
@@ -19,13 +19,12 @@ import argparse
 import logging
 
 import jax.random as random
-from jax.random import PRNGKeyArray
 import json
 import os
 
 import pyggdrasil.tree_inference as tree_inf
 
-from pyggdrasil.tree_inference import CellSimulationModel
+from pyggdrasil.tree_inference import CellSimulationModel, CellSimulationId, TreeId
 from pyggdrasil.serialize import JnpEncoder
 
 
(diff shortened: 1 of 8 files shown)


gordonkoehn:
This code is untested as it depends on 
#63 
but will be tested and bound in the near future.

scripts/cell_simulation.py
```diff
@@ -106,58 +109,61 @@ def create_parser() -> argparse.Namespace:
...
+        strategy=params.strategy,
+        fpr=params.fpr,
+        fnr=params.fnr,
+        na_rate=params.na_rate,
+        observe_homozygous=params.observe_homozygous,
+    )
+
+    # set up the CellSimulationModel from the params and Tree Information from TreeId
     params_dict = vars(params)
+    params_dict["n_mutations"] = tree_id.n_nodes - 1
```
gordonkoehn, line 151: Note how the `n_mutations` parameter is now defined by the true trees filename i.e. `TreeId`.  Just one wildcard in the snakemake to rule both

src/pyggdrasil/tree_inference/_file_id.py
```diff
@@ -94,6 +94,23 @@ def _create_id(self) -> str:
...
+        # split string by underscore and assign to attributes
+        _, tree_type, n_nodes, seed = str_id.split("_")
+
+        if seed is not None:
+            tree_id = TreeId(TreeType(tree_type), int(n_nodes), int(seed))
+            return tree_id
+        else:
+            tree_id = TreeId(TreeType(tree_type), int(n_nodes))
+            return tree_id
+
```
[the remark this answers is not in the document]
pawel-czyz: I've seen people doing this, but I'm personally not a fan since I learned Snakemake...

src/pyggdrasil/tree_inference/_tree_generator.py
```diff
@@ -42,7 +43,7 @@ def generate_star_Tree(n_nodes: int) -> Tree:
     Root node is the highest index node."""
     adj_mat = _generate_star_tree_adj_mat(n_nodes)
     labels = np.arange(n_nodes)
-    return Tree(jax.Array(adj_mat), jax.Array(labels))
+    return Tree(jnp.array(adj_mat), jnp.array(labels))
```
pawel-czyz, line 46: Good catch!

pawel-czyz (approved):
Nice! I'm not sure if encoding the parameters in file IDs (with serialization/deserialization to the filename) is the best way (as we could do this by having Pydantic objects serialized to JSON or use Snakemake), but if you think it's convenient to work with, we can keep it.

scripts/cell_simulation.py
gordonkoehn: Mmhh - alright - I think I still don't quite understand your method -  how to expand the dictionary `scenario` way to so many different parameters without having very long dictionaries.

src/pyggdrasil/tree_inference/_file_id.py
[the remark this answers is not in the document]
gordonkoehn: ok -  I could pass all the arguments instead of the filename but avoids optional parameters. - not quite sure how to do this with snakemake only.

Issue #60, opened 2023-06-07
Write Script to make huntress initial tree from cell simulation data


gordonkoehn:
Given a mutation matrix generated from the cell simulation data or and other mutation matrix data for that matter, there should be a script to generate the initial Tree using Huntress.

PR #68, opened 2023-06-08
Write Script for Huntress Tree Generation

gordonkoehn:
This PR addresses
- #60

### DIFF
--- a/scripts/make_huntress.py
+++ b/scripts/make_huntress.py
@@ -27,17 +27,19 @@ from pyggdrasil.tree_inference import (
     huntress_tree_inference,
     CellSimulationId,
     MutationDataId,
+    TreeId,
+    TreeType,
 )
 
 
 #############################################
 # Placeholder Fns implemented in PR #42
 
-from typing import TypedDict
 
+class CellSimulationData:
+    """TypedDict representing the data output of cell_simulation.py"""
 
-class CellSimulationData(TypedDict):
-    pass
+    raise NotImplementedError
 
 
 def get_simulation_data(data: dict):
@@ -57,7 +59,7 @@ def get_simulation_data(data: dict):
             root: TreeNode
                 Root of the tree.
     """
-    pass
+    raise NotImplementedError
 
 
 #############################################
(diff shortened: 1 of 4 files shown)


scripts/make_tree.py
```diff
@@ -22,29 +22,13 @@
...
-
-
-#####################################################
+from pyggdrasil.tree_inference import (
+    TreeId,
+    TreeType,
+    generate_random_TreeNode,
+    generate_star_TreeNode,
+    generate_deep_TreeNode,
+)
```
gordonkoehn, line 31: Unrelated to this PR - these changes are just removing a dead reference to placeholder functions now already merged.

scripts/make_huntress.py
```diff
@@ -0,0 +1,175 @@
...
+    args = parser.parse_args()
+
+    return args
+
+
+def main() -> None:
+    """
+    Main function for plotting trees from MCMC samples.
+
+    Returns:
```
pawel-czyz, line 121: In general, I don't think `Returns None` in the docstring is needed, as it can be read from the type hint :wink:

scripts/make_huntress.py
```diff
@@ -0,0 +1,175 @@
...
+        # get the cell simulation id from the filename
+        CellSimulationId.from_str(filename)
+
+    except AssertionError:
+        # TODO: consider adding logger
+        print("Could not get CellSimulationId from filename, switching to MutationID")
+        # make mutation id instead
+        MutationDataId(filename)
+
+    except Exception as E:
```
pawel-czyz, line 160: I think you can do just
```
except Exception:
    raise
```
and it'll raise the last exception on the stack.

scripts/make_tree.py
pawel-czyz: Good idea!

src/pyggdrasil/tree_inference/_file_id.py
```diff
@@ -156,6 +172,52 @@ def _create_id(self) -> str:
     def __str__(self) -> str:
         return self.id
 
+    @classmethod
+    def from_str(cls, str_id: str):
+        """Creates a CellSimulation id from a string representation of the id.
+        Args:
+            str_id: str
+
+        Throws:
```
pawel-czyz, line 181: I don't think a change needed here, but in the future I'd recommend `Raises` rather than `Throws` (so we can use automated documentation building utilities using Google style comments).

src/pyggdrasil/tree_inference/_file_id.py
```diff
@@ -156,6 +172,52 @@ def _create_id(self) -> str:
...
+        """
+        # split string by underscore and assign to attributes
+        # CS_1-T_d_10-100_0.01_0.01_0.01_true_UXR
+        cs_par1, tree_id, cs_part2 = str_id.split("-")
+        # check prefix and postfix
+        assert cs_par1.startswith("CS_")
+        assert tree_id.startswith("T_")
+        assert cs_part2.endswith("UXR") or cs_part2.endswith("UIR")
+        # split cs_part2 by underscore and assign to attributes
+        cs_part2 = cs_part2.split("_")
```
pawel-czyz, line 192: This logic is very complex – at least a little amount of unit tests acting both as sanity checks and examples what to expect would be helpful in the long run.

pawel-czyz (approved):
I don't really understand the logic inside `_file_id.py`, but otherwise looks good to me!

src/pyggdrasil/tree_inference/_file_id.py
gordonkoehn: Added one - caught mistakes actually. Good call !

Issue #66, opened 2023-06-08
Revise MCMC Script


gordonkoehn:
Resise the `run_mcmc.py` script to 
- [x] use the new file naming of #54
- [x] use trees generated by #61 
- [x] #69

Ideally, done after 
- #23 
is done, to keep PRs small Some changes in `run_mcmc.py`

gordonkoehn:
Regarding removing the MCMC run config file -  @pawel-czyz may I run this setup by you.

- create a `pydantic` model for the  MCMC run configs - import the ones that already exist in `pyggdrasil` 
- set te `pydantic` models in the snakemake 
- pass the models as a dict - i.e. serialisable to a the shell running the python script

This allows me to use a `pydantic` to check the configurations as well as keeping the code to run a mcmc simulation in a script -  as it contains a lot of logging and file handeling better to keep outside of `pyggdrasil`.

So effectively I am proposing a snakemake as this: 

```
import json
from pydantic import BaseModel
from snakemake import shell

### import from pyggdrasil
# Define a Pydantic model
class MoveProbs(BaseModel):
    prune_and_reattach: float
    swap_node_labels: float
    swap_subtrees: float

class Config(BaseModel):
    move_probs: MoveProbs
    fnr: float
    fpr: float
    num_samples: int
    burn_in: int
    thinning: int
####

# Set the Model
# Create an instance of the Pydantic model with hardcoded values
config = Config(
    move_probs=MoveProbs(
        prune_and_reattach=0.1,
        swap_node_labels=0.65,
        swap_subtrees=0.25
    ),
    fnr=0.097,
    fpr=1.24e-06,
    num_samples=100,
    burn_in=10,
    thinning=2
)

rule my_rule:
    output:
        "output.txt"
    shell:
        "python my_script.py {output} {params.config_dict_json}"
    
params:
    config_dict_json=json.dumps(config.dict())
```

pawel-czyz:
Hi Gordon,

This should work!

An alternative:

```
from cool_package import run_some_function
...

CONFIGS = {
   "realistic_config": Config(
    move_probs=MoveProbs(
        prune_and_reattach=0.1,
        swap_node_labels=0.65,
        swap_subtrees=0.25
    ),
    fnr=0.097,
    fpr=1.24e-06,
    num_samples=100,
    burn_in=10,
    thinning=2
)

rule all:
   input: expand("output-{scenario}.txt", scenario=CONFIGS.keys())

rule my_rule:
[...]

PR #80, opened 2023-06-15
Revise MCMC Script 2

gordonkoehn:
This PR addresses 
- #66 
and contributed partly to 
- #81

## Contents of this PR
The `run_mcmc.py` script is updated to 
- implement the new `McmcRunId` and alike filenames
- the snakemake rules from tree generation to mcmc run work together by plain string matching.

This required some minor changes in `serialize` and `tree_inference`. 

I removed a lot of unnecessary functionality as
- restarting the mcmc run for the output directly (rather use a `get_tree.py` to extract the restart tree from the mcmc samples)
- remove timestamps in filenames
- make output of cell simulation static (always output noisy matrix - irrespective of error rates)
- remove ad-hoc tree generation of mcmc chain

## Comment on Attempts to move away from running scripts via `shell`
I looked into a way to run the mcmc sampler from the snakemake, in other ways than by using the shell. 
Without rewriting all scripts as functions to call in the `python` command, I tried to use the `run` command and pass the arguments as `snakemake.input[0]`. This would have been nice. Yet, unfortunately, this would require installing snakemake in poetry which does not work. So for not I'll stick to this plain script / string matching of snakemake.

## Future Directions
I indent to use the enum options I built for MCMC configs, to be used in later snakemake rules specific for experiments. 
The current snakemake rules are able to handle all possible configurations. - This should remove a lot of parameters when specifying experiments.

### DIFF
--- a/scripts/run_mcmc.py
+++ b/scripts/run_mcmc.py
@@ -177,8 +177,13 @@ def run_chain(
     # Set random seed
     rng = random.PRNGKey(params.seed)
 
-    # load observed mutation matrix data from file
-    mut_mat = jnp.array(get_mutation_matrix(params.data_fp))
+    # load data of mutation matrix
+    with open(params.data_fp, "r") as f:
+        cell_simulation_data = json.load(f)
+    cell_simulation_data = tree_inf.get_simulation_data(cell_simulation_data)
+    # get the mutation matrix
+    # TODO: adjust to flexible noisy / perfect
+    mut_mat = cell_simulation_data["perfect_mutation_mat"]
 
     # check if init tree is provided
     if params.init_tree_fp is None:
--- a/scripts/run_mcmc.py
+++ b/scripts/run_mcmc.py
@@ -40,22 +40,23 @@ JSON tree_inference.McmcConfig
 import argparse
 import jax.random as random
 import json
-import jax.numpy as jnp
 import logging
 import pytz
 
 from pathlib import Path
 from datetime import datetime, timedelta
-from typing import Optional
 
 import pyggdrasil.tree_inference as tree_inf
 import pyggdrasil.serialize as serialize
 
 from pyggdrasil.tree_inference import (
-    MutationMatrix,
     mcmc_sampler,
     MoveProbabilities,
     McmcConfig,
+    TreeId,
+    McmcRunId,
+    CellSimulationId,
+    MutationDataId,
 )
 
 
@@ -83,27 +84,8 @@ def create_parser() -> argparse.Namespace:
     )
 
     parser.add_argument(
-        "--init_tree_mcmc_no",
-        required=False,
-        help="Sample number if mcmc sample is provided as initial tree. "
-        "- line number in file from 1",
-        default=None,
-        type=int,
-    )
-
-    parser.add_argument(
-        "--init_TreeNode",
-        required=False,
-        help="Sample number if mcmc sample is provided as initial tree.",
-        default=None,
-        action="store_true",
-    )
-
-    parser.add_argument(
         "--config_fp", required=False, help="Config file path", type=str
     )
-
-    parser.add_argument("--config", required=False, help="Raw config in json", type=str)
 
     parser.add_argument(
         "--out_dir",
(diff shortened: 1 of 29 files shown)


gordonkoehn:
Calling the snakemake with a file specific command for an MCMC output file as
` snakemake -c 4 ../data/mark00/mcmc/MCMC_42-CS_42-T_d_10_123-1000_0.01_0.02_0.0_f_UXR-iT_d_10_123-MC_1.24e-06_0.097_1000_0_1-MPC_0.1_0.65_0.25.json
`
Runs all these rules:
Screenshot 2023-06-16 at 12 46 27

And results in all these files: 
Screenshot 2023-06-16 at 12 45 45

scripts/run_mcmc.py
```diff
@@ -86,29 +83,10 @@ def create_parser() -> argparse.Namespace:
...
-        default=None,
-        action="store_true",
-    )
-
     parser.add_argument(
         "--config_fp", required=False, help="Config file path", type=str
     )
 
-    parser.add_argument("--config", required=False, help="Raw config in json", type=str)
-
```
gordonkoehn, line 111: removing arguments - in general streamlining the inputs and functionality of this script

scripts/run_mcmc.py
```diff
@@ -150,52 +128,18 @@ def run_chain(
...
+    # parse tree from file given as input
+    # make path and check if Path exists
+    p = Path(params.init_tree_fp)
+    if not p.exists():
+        raise FileNotFoundError(f"File {params.init_tree_fp} does not exist.")
+
+    init_tree_node = serialize.read_tree_node(params.init_tree_fp)
+    # convert TreeNode to Tree
+    init_tree = tree_inf.tree_from_tree_node(init_tree_node)
+    logging.info("Loaded tree (TreeNode) from file.")
```
gordonkoehn, line 142: much simpler with less optional parameters :)

scripts/run_mcmc.py
```diff
@@ -268,27 +211,48 @@ def get_str_timedelta(td: timedelta) -> str:
...
+    # get data id
+    # get data id from filepath excluding json extension
+    data_path = Path(params.data_fp)
+    data_id = data_path.stem
+    try:
+        data_id = CellSimulationId.from_str(data_id)
+    except AssertionError:
+        data_id = MutationDataId(data_id)
+    # make MC run id
+    mc_run_id = McmcRunId(params.seed, data_id, init_tree_id, config_id)
```
gordonkoehn, line 235: Note - all namings should be able to handle `mutation data` of arbitrary names :) - if we ever switch to real data.

src/pyggdrasil/tree_inference/_simulate.py
```diff
@@ -672,19 +670,13 @@ def gen_sim_data(
...
-            "perfect_mutation_mat": perfect_mutation_mat.tolist(),
-            "root": root_serialized,
-        }
+
+    data = {
+        "adjacency_matrix": tree_adj_mat.tolist(),
+        "perfect_mutation_mat": perfect_mutation_mat.tolist(),
+        "noisy_mutation_mat": noisy_mutation_mat.tolist(),
+        "root": root_serialized,
+    }
```
gordonkoehn, line 679: Made the output of `cell_simulation.py` static. - Allways output the same dictionary with the `noisy_mutation_mat' even if noise is none. - Makes this a lot easier to handle.

workflows/Snakefile
```diff
@@ -13,33 +17,33 @@ params_md["fnr"] = 0.097
...
+params_mcmc["seed"] = 42
+params_mcmc["n_trees"] = 1
+params_mcmc["n_cells"] = 2000
+params_mcmc["n_mutations"] = 50
+params_mcmc["strategy"] = """UNIFORM_INCLUDE_ROOT"""
+params_mcmc["fpr"] = 1.24*10**(-6)
+params_mcmc["fnr"] = 0.097
+params_mcmc["na_rate"] = 0.014
+params_mcmc["observe_homozygous"] = "False"
+
```
gordonkoehn, line 31: params not in use yet - will come later once I define experiments - current snakemake rules can handle all possible parameters

workflows/Snakefile
```diff
@@ -13,33 +17,33 @@ params_md["fnr"] = 0.097
...
-                            fpr=params_md["fpr"],
-                            fnr=params_md["fnr"],
-                            na_rate=params_md["na_rate"],
-                            observe_homozygous=params_md["observe_homozygous"],
-                            tree_no= list(range(1, params_md["n_tree"]+1))
-                        ),
-        mcmc_log_mark0ß = '../data/mcmc/mark00/mcmc_run.log',
-        mcmc_samples_mark00 = '../data/mcmc/mark00/samples.json'
-
-
```
gordonkoehn, line 35: removing the old

workflows/Snakefile
```diff
@@ -88,46 +92,96 @@ rule gen_cell_simulation:
...
+    run:
+        # define the move probabilities manually
+        custom_MP_conf = MoveProbConfig(
+            prune_and_reattach=float(params.prune_and_reattach),
+            swap_node_labels=float(params.swap_node_labels),
+            swap_subtrees=float(params.swap_subtrees)
+        )
+
+        with open(output.mcmc_move_prob_config, "w") as f:
+            json.dump(custom_MP_conf.dict(), f)
```
gordonkoehn, line 123: Didn't manage to get pydantic objects / Json to be passed to my scripts whilst using simple string matching of snakemake with my filenames.

So my alternative is: have rules that create my configurations pydantic objects using the snakemake wildcards and save those to disk - then to be read in by the later rules. 
- ensures correct typing of my configurations
- allows to use pure string matching of fully informative filenames

scripts/run_mcmc.py
pawel-czyz: Nice! Also, more compatible with single responsibility principle.

scripts/run_mcmc.py
pawel-czyz: Good!

src/pyggdrasil/tree_inference/_simulate.py
pawel-czyz: This is an excellent idea!

workflows/Snakefile
```diff
@@ -13,33 +17,33 @@ params_md["fnr"] = 0.097
...
-
-
+        mcmc_samples = "Test.json"
 
 rule make_random_or_deep_tree:
     """Make a tree (TreeNode) of random/deep topology and save it as JSON."""
     input:
         script="../scripts/make_tree.py"
+
+    wildcard_constraints:
```
pawel-czyz, line 43: I thought that `expand` for the experimental design would actually be able to understand constraints, but I may be wrong

workflows/Snakefile
pawel-czyz: Agree, this should work :slightly_smiling_face:  Also, I still very much like passing them straight to the rules using `run:` rather than `script:`

pawel-czyz (approved):
LGTM!

workflows/Snakefile
gordonkoehn: Oh, my idea is to use expand later for my experimental specification. All these rules you see here provide snakemake the most general way to produce all files.

Issue #83, opened 2023-06-19
Investigating Bug - Tree breaking apart 2


gordonkoehn:
During the #82, some of the deep trees in the mcmc run generated did not have a root anymore. 

Something went wrong again. 

Need to investigate what is happening here.

PR #82, opened 2023-06-16
Implement Distance Utilities

gordonkoehn:
Addresses #74

Issue #14, opened 2023-04-14
Add Tests for MCMC Tree Moves


gordonkoehn:
Write unit tests to ensure that the MCMC moves work as expected. Functions to test:

`_mcmc._swap_node_labels_proposal`
             `._prune_and_reattach_proposal`
             `._swap_subtrees_proposal`
             
             
Several testing strategies: (Proposed by @pawel-czyz in #12)

1. For _swap_node_labels_proposal: if we have root + 3 nodes and sample 1,000 times, will we arrive at (approximately) 1000/3! samples in each category? Is the root always preserved to be the last one?
2. For _prune_and_reattach and swap_subtrees_proposal: do we always end up with a cycle-less tree? Is the root never pruned/swapped?

PR #84, opened 2023-06-19
Fixing Tree Move Bugs #2

gordonkoehn:
This PR addresses 
- #83 

During the development of 
- #82 
some trees generated in the mcmc chain of an initial deep tree could again not be converted to `TreeNodes`, no root was there. The tree was broken. 

There must be a bug in the mcmc moves. After already having fixed one in 
- #48 
and running chains of hundreds of samples it is surprising to see this. Glad to find the bugs now. 
It appears small deep trees test the extreme cases of the tree moves.

So far I have found 2 bugs. One in 
- `_prune_and_reattach_proposal` : the pruned node itself was included in the potential nodes to reattach to
-`_swap_node_labels_move` : confused node levels with indices

I have lost trust in these moves, I have already added test functions, but it is about time to test these moves thoroughly. I'll add tests as proposed in, i.e. this PR also closes
- #14

### DIFF
--- a/src/pyggdrasil/tree_inference/_mcmc.py
+++ b/src/pyggdrasil/tree_inference/_mcmc.py
@@ -96,15 +96,16 @@ def _prune_and_reattach_proposal(key: JAXRandomKey, tree: Tree) -> Tuple[Tree, f
 def _swap_node_labels_move(tree: Tree, node1: int, node2: int) -> Tree:
     """Swaps labels between ``node1`` and ``node2`` leaving the tree topology
     untouched."""
-    label1 = tree.labels[node1]
-    label2 = tree.labels[node2]
+
+    node1_idx = jnp.where(tree.labels == node1)[0]
+    node2_idx = jnp.where(tree.labels == node2)[0]
 
     # Copy all the labels
     new_labels = tree.labels
     # Assign label2 to node1...
-    new_labels = new_labels.at[node1].set(label2)
+    new_labels = new_labels.at[node1_idx].set(node2)
     # ... and label1 to node2
-    new_labels = new_labels.at[node2].set(label1)
+    new_labels = new_labels.at[node2_idx].set(node1)
 
     logger.debug(
         "MCMC: Swap node labels move - swapped labels of node %s and node %s",
--- a/src/pyggdrasil/tree_inference/_mcmc.py
+++ b/src/pyggdrasil/tree_inference/_mcmc.py
@@ -79,10 +79,23 @@ def _prune_and_reattach_proposal(key: JAXRandomKey, tree: Tree) -> Tuple[Tree, f
     rng_prune, rng_reattach = random.split(key)
     # pick a random non-root node to prune
     pruned_node = int(random.choice(rng_prune, tree.labels[:-1]))
+
+    # TODO (Gordon) : consider to remove if checks passed
+    logger.debug("MCMC: Prune and reattach move - pruned node %s", pruned_node)
+
     # get descendants of pruned node
-    descendants = get_descendants(tree.tree_topology, tree.labels, pruned_node)
-    # possible nodes to reattach to - including pruned node for aperiodic case
+    descendants = get_descendants(
+        tree.tree_topology, tree.labels, pruned_node, include_parent=True
+    )
+    # possible nodes to reattach to - no descendants of pruned node,
+    # nor pruned node itself, but root
     possible_nodes = jnp.setdiff1d(tree.labels, descendants)
+
+    # TODO (Gordon) : consider to remove if checks passed
+    logger.debug(
+        "MCMC: Prune and reattach move - possible nodes to reattach to: %s",
+        possible_nodes,
+    )
     # pick a random node to reattach to
     attach_to = int(random.choice(rng_reattach, possible_nodes))
     return (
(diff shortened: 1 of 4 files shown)


src/pyggdrasil/tree_inference/_tree.py
```diff
@@ -68,6 +69,14 @@ def print_topo(self):
...
         return self.to_TreeNode().print_topo()
 
+    def __str__(self):
+        labels = self.labels
+        assert labels is not None
+        df = pd.DataFrame(
+            self.tree_topology.astype(int), labels, labels  # type: ignore
+        )
+        return df.__str__()
+
```
gordonkoehn, line 79: Easier Debugging

gordonkoehn (commented):
Comments on bug fix 1 and 2

tests/tree_inference/test_mcmc.py
```diff
@@ -349,3 +350,131 @@ def test_swap_subtrees_move_fig17_nested_subtrees(seed):
...
+        """Log prob function for testing. - dummy function"""
+        return random.uniform(rng, shape=()).__float__()
+
+    # Run the kernel
+    for i in range(n_moves):
+        rng, rng_now = random.split(rng)
+        tree, _ = mcmc._mcmc_kernel(
+            rng_now, tree, move_probs, log_prob_fn  # type: ignore
+        )
+        assert tr.is_valid_tree(tree)
```
gordonkoehn, line 480: Note this test - runs a mock mcmc with a few samples and checks that the tree is valid still. 
This is the best real test I can think of.

tests/tree_inference/test_tree.py
```diff
@@ -221,3 +225,32 @@ def test_is_not_same_tree():
...
+
+
+@pytest.mark.parametrize("n", [5, 10])
+@pytest.mark.parametrize("seed", [34, 424])
+@pytest.mark.parametrize("tree_type", ["r", "s", "d"])
+def test_is_valid_tree(n: int, seed: int, tree_type: str):
+    """Test is_valid_tree function."""
+
+    tree = make_tree(n, seed, tree_type)
+    assert tr.is_valid_tree(tree) is True
```
gordonkoehn, line 256: may look circular - yet I am sure that the tree generator functions are working fine

src/pyggdrasil/tree_inference/_mcmc.py
```diff
@@ -79,10 +79,23 @@ def _prune_and_reattach_proposal(key: JAXRandomKey, tree: Tree) -> Tuple[Tree, f
...
+
+    # TODO (Gordon) : consider to remove if checks passed
+    logger.debug("MCMC: Prune and reattach move - pruned node %s", pruned_node)
+
     # get descendants of pruned node
-    descendants = get_descendants(tree.tree_topology, tree.labels, pruned_node)
-    # possible nodes to reattach to - including pruned node for aperiodic case
+    descendants = get_descendants(
+        tree.tree_topology, tree.labels, pruned_node, include_parent=True
+    )
```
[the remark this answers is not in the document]
pawel-czyz: Wow, this was tricky. Good you found it out!

src/pyggdrasil/tree_inference/_mcmc.py
```diff
@@ -96,15 +109,16 @@ def _prune_and_reattach_proposal(key: JAXRandomKey, tree: Tree) -> Tuple[Tree, f
...
+    node2_idx = jnp.where(tree.labels == node2)[0]
 
     # Copy all the labels
     new_labels = tree.labels
     # Assign label2 to node1...
-    new_labels = new_labels.at[node1].set(label2)
+    new_labels = new_labels.at[node1_idx].set(node2)
     # ... and label1 to node2
-    new_labels = new_labels.at[node2].set(label1)
+    new_labels = new_labels.at[node2_idx].set(node1)
```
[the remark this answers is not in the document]
pawel-czyz: Right. It may be good to check for this sneaky bug also in the other two moves 
 (permuting the labels and swapping subtrees).

src/pyggdrasil/tree_inference/_tree.py
```diff
@@ -310,3 +321,31 @@ def is_same_tree(tree1: Tree, tree2: Tree) -> bool:
         ) and jnp.all(tree1.labels == tree2_reordered.labels)
 
     return bool(result)
+
+
+def is_valid_tree(tree: Tree) -> bool:
```
pawel-czyz, line 326: Good idea, it should catch some bugs. Do we know that `to_TreeNode` explicitly check for these four things or we hope it does it implicitly?

tests/tree_inference/test_mcmc.py
pawel-czyz: It's a good one :slightly_smiling_face:

pawel-czyz (approved):
Very nice you found it! I'd suggest to check for the label/label index bug also in other tree moves, as it's a very sneaky one...

src/pyggdrasil/tree_inference/_tree.py
gordonkoehn: Yes, converting to `TreeNode` from `Tree` is what so far has caught all the broken trees. To convert it needs to check for the 1) root, 2) self-loops would lead to recursion failing 3)  not connected components leads to multiple roots, also caught

src/pyggdrasil/tree_inference/_mcmc.py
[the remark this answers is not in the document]
gordonkoehn: Yes, they are tested now. 

`swap_subtrees` is correct since #49 

and prune and reattach was also fixed here.

PR #85, opened 2023-06-20
Gordon/implement distance utilities rebase

gordonkoehn:


gordonkoehn:
a failed attempt to create a cleaner diff view after merger of feature into feature branch

Issue #89, opened 2023-06-22
Division by Zero in scphylo ancestor-descendant distance


gordonkoehn:
For the below `mark00` configuration with `mutations = 4` `scphylo` did give a `DivisionByZero` error. 

Larger trees don't get this error. 

![Screenshot 2023-06-22 at 12 46 51](https://github.com/cbg-ethz/PYggdrasil/assets/10867778/7edda533-a58a-4805-845c-aecc41aeec5c)

gordonkoehn:
Suggest to `try catch` the calculation and output the very tree this happens for.

Issue #73, opened 2023-06-12
MARK01: How well does HUNTRESS work?


gordonkoehn:
## Subtasks
- [x] #91
- [x] #92
- [x] #93

## Description Product

![Image](https://user-images.githubusercontent.com/10867778/245129217-f0f502e8-503e-4b9c-a07d-de198f2dbac4.png)

_Does HUNTRESS really work: generate histograms of distances/similarities for different:_

FPR and FNR (ideal, average, bad)
Mutations: 5, 10, 30, 50
Number of cells: 200, 1000, 5000

x-axis: distance to true tree
y-axis: freq over independent runs of huntress of
fixed tree and 200 data sets with sampled cells from that tree. For each data set we know that the tree is the same, but HUNTRESS may return a different one 
200 different trees, each tree with a data set. Measure the distance between the true tree and the inferred one

Outcome:
	if works: great, SCITE is faster
	if not works: Huntress may not be the way to go, but perhaps still better than random

Note: may use MCMC edit distance later

gordonkoehn:
- [ ] Check if number of cells higher than mutations -  that's when HUNTRESS claims to be good.

gordonkoehn:
@pawel-czyz What do we choose as a true tree here ? - Just a random tree I suppose ?

gordonkoehn:
@pawel-czyz What distances shall we go for: MP3 and AD?

PR #90, opened 2023-06-22
Setting up experiment MARK01

gordonkoehn:
Implementing the experimental workflow of 
- #73 

Produces this massive DAG:
[mark01_dag.pdf](https://github.com/cbg-ethz/PYggdrasil/files/11881828/mark01_dag.pdf)

This PR closes:
- #91 
- #92 

Before we can run the experiment it is worth investigating why the scyphylo AD distance still fails sometimes. 
- #89


src/pyggdrasil/serialize/_to_json.py
```diff
@@ -207,9 +212,12 @@ def save_metric_result(iteration: list[int], result: list[float], out_fp: Path)
...
+def read_metric_result(fullpath: Path) -> tuple[list[Union[int, str]], list[float]]:
     """Reads in all metric results from JSON file for a given run."""
     with open(fullpath, "r") as f:
         metric_dict = json.load(f)
 
-    return metric_dict["iteration"], metric_dict["result"]
+    # get axis name
+    axis_name = list(metric_dict.keys())[0]
+
+    return metric_dict[axis_name], metric_dict["result"]
```
gordonkoehn, line 223: Reusing these functions - made them more flexible.

src/pyggdrasil/tree.py
```diff
@@ -56,5 +56,21 @@ def print_topo(self) -> None:
...
+        Args:
+              node: AnyTree node to be converted
+        Returns:
+            TreeNode with the same topology as the AnyTree node
+        """
+        children = [
+            TreeNode.convert_anytree_to_treenode(child) for child in node.children
+        ]
+        return TreeNode(name=node.name, children=children)
+
```
gordonkoehn, line 74: Tree Conversions still haunt me :) 

Hope this was the last one.

src/pyggdrasil/tree_inference/_file_id.py
```diff
@@ -208,15 +208,25 @@ def from_str(cls, str_id: str):
         """
         # split string by underscore and assign to attributes
         # CS_1-T_d_10-100_0.01_0.01_0.01_true_UXR
-        cs_par1, tree_id, cs_part2 = str_id.split("-")
+        parts = str_id.split("-")
+        cs_part1 = parts[0]
+        tree_id = parts[1]
+        print(tree_id)
+        cs_part2 = "-".join(parts[2:])
+
```
gordonkoehn, line 216: Allowing for scientific notation error rates. 

The trouble with string matching.

workflows/mark01.smk
```diff
@@ -0,0 +1,160 @@
...
+# Homozygous mutations [f: False / t: True]
+observe_homozygous = "f" # <-- configure whether to observe homozygous mutations here
+
+# cell attachment strategy [UXR: Uniform Exclude Root / UIR: Uniform Include Root]
+cell_attachment_strategy = "UXR" # <-- configure cell attachment strategy here
+
+#####################
+# True Tree Parameters
+tree_types = ["r"] # <-- configure tree type here ["r","s","d"]
+tree_seeds = [42] # <-- configure tree seed here
```
gordonkoehn, line 44: Easy to choose more than one true tree.

workflows/mark01.smk
```diff
@@ -0,0 +1,160 @@
...
+    run:
+        import pyggdrasil as yg
+        from pathlib import Path
+        import matplotlib.pyplot as plt
+        # load the distances
+        tree_ids, distances =  yg.serialize.read_metric_result(Path(input.distances))
+        # make the histogram
+        fig, axs = plt.subplots(1,1,tight_layout=True)
+        # We can set the number of bins with the *bins* keyword argument.
+        axs.hist(distances, bins='auto')
```
gordonkoehn, line 94: A preference for binning? Currently using the "auto = "`Maximum of the ‘sturges’ and ‘fd’ estimators. `

gordonkoehn: https://numpy.org/doc/stable/reference/generated/numpy.histogram_bin_edges.html#numpy.histogram_bin_edges

gordonkoehn (commented):
Full MARK01 workflow - fix AD distance next. 

So far I have not seen any interesting histograms during testing. 

Huntress either gave a tree of MP3 = 0 or 1 for all setups. 

Hope more exciting is to come.

workflows/mark01.smk
pawel-czyz: Any choice is good, I think – we'll probably want to load the data and play with the bin settings in a notebook anyway (or look at other visualisations in more exploratory manner)

tests/tree_inference/test_file_id.py
```diff
@@ -105,3 +122,26 @@ def test_cell_simulation_id_from_str(cell_simulation_id) -> None:
     assert csi.na_rate == cell_simulation_id.na_rate
     assert csi.observe_homozygous == cell_simulation_id.observe_homozygous
     assert csi.strategy == cell_simulation_id.strategy
+
+
+def test_cell_simulation_id_from_str_scientific_notation(
```
pawel-czyz, line 127: Nice!

pawel-czyz (approved):
LGTM!

src/pyggdrasil/tree_inference/_file_id.py
gordonkoehn: Yes, I completely see how one could do this now! Understand Snakemake better now. Would do things differently now :)  Would have made the output file names more human-readable. 

Would be quite a bit of refactoring in `workflows`, `tests` and `tree_inference` to change this now.

I don't see any danger of the error rates representation as numbers breaking down now. 

I'll stick to the current solution. This is probably something I won't change after running the first experiments.

If final output files get tiring to read I write a rule to rename files for ease of finding the files one wants.

PR #8, opened 2023-03-09
Add installation instructions

pawel-czyz:
This PR adjusts the ReadMe, explaining the tools and conventions configured for this repository. What do you think?

PR #100, opened 2023-06-28
Resolving broken `scphylo` ancestor-descendant distance

gordonkoehn:
This PR addresses 
- #89 

To investigate why the Ancestor-Descendant (AD) distance currently in use from our `scphylo` fork does not work, I implemented tests for both MP3 and AD distance. 

The AD distance fails in a pattern I have not yet understood, and if it works its result differs from manual calculation. See my calculation in #89. 

The three issues I can think of are. 
1) `scphylo` AD is broken
2) we use `scphylo` wrongly - wrong input
3) different definitions of AD

2 : The MP3 distances of `scphylo` appear sensible - whilst I have not manually checked them, because that takes a little more effort.  Further, I double checked the inputs. - Unlikely that it is this one. 
3: checked docstring seems the same - same as in B-SCITE

This leads me to the conjecture that `scphylo` AD is broken. 

Shall I open a well documented issue in their repo ?

In the meanwhile / or instead we can use Laura's version of the AD - works like a charm. 

Laura was so kind to share here code for the AD - @laurabquintas thanks a lot !

Made it de default for the AD in this PR.


src/pyggdrasil/analyze/_metrics.py
```diff
@@ -15,7 +15,7 @@ def get(metric: str) -> Callable[[TreeNode, TreeNode], float]:
         return Metrics._METRICS[metric]
 
     _METRICS: Dict[str, Callable[[TreeNode, TreeNode], float]] = {  # type: ignore
-        "AD": AncestorDescendantSimilarity().calculate,
+        "AD": AncestorDescendantSimilarity_lq().calculate,
```
gordonkoehn, line 18: Making Laura's the default -  refactoring the `*_lq` away once we know if `scyphylo` is indeed broken

src/pyggdrasil/distances/_similarities.py
```diff
@@ -0,0 +1,45 @@
...
+            for node in anytree.PreOrderIter(tree):
+                if not node.is_leaf:
+                    for child in node.descendants:
+                        pairs.add((node.name, child.name))
+            return pairs
+
+        pairs1 = create_pairs(tree1)
+        pairs2 = create_pairs(tree2)
+
+        return len(pairs1.intersection(pairs2)) / len(pairs1.union(pairs2))
```
gordonkoehn, line 35: @laurabquintas Thanks so much! - here is your code :)

tests/distances/test_scphylo_wrapper.py
```diff
@@ -63,3 +70,164 @@ def test_another_is_different_and_symmetric(similarity, model_tree) -> None:
...
+    def tree1_(n_nodes: int) -> TreeNode:
+        """Return a tree."""
+        if tree_type == "r":
+            return ti.generate_random_TreeNode(rng, n_nodes)
+        elif tree_type == "d":
+            return ti.generate_deep_TreeNode(rng, n_nodes)
+        else:  # tree_type == "s"
+            return ti.generate_star_TreeNode(n_nodes)
+
+    return tree1_
```
gordonkoehn, line 88: Just playing around with currying ;)

tests/distances/test_scphylo_wrapper.py
```diff
@@ -63,3 +70,164 @@ def test_another_is_different_and_symmetric(similarity, model_tree) -> None:
...
+@pytest.mark.skip("scyphylo metric under investigation")
+@pytest.mark.parametrize("tree_type1", ["r", "d", "s"])
+@pytest.mark.parametrize("seed1", [76, 42])
+@pytest.mark.parametrize("tree_type2", ["r", "d", "s"])
+@pytest.mark.parametrize("seed2", [13, 42])
+@pytest.mark.parametrize("n_nodes", [5, 10])
+def test_AncestorDescendantSimilarity(
+    n_nodes: int, tree_type1, seed1: int, tree_type2, seed2: int
+):
+    """Test the AncestorDescendantSimilarity scyphylo class."""
```
gordonkoehn, line 133: Failing test - `scyphylo` just throws `DivisionByZero` errors, every about 5 trees.

tests/distances/test_scphylo_wrapper.py
```diff
@@ -63,3 +70,164 @@ def test_another_is_different_and_symmetric(similarity, model_tree) -> None:
...
+@pytest.mark.parametrize("tree_type1", ["r", "d", "s"])
+@pytest.mark.parametrize("seed1", [76, 42])
+@pytest.mark.parametrize("tree_type2", ["r", "d", "s"])
+@pytest.mark.parametrize("seed2", [13, 42])
+@pytest.mark.parametrize("n_nodes", [5, 10])
+def test_AncestorDescendantSimilarity_scyphylo_lq(
+    n_nodes: int, tree_type1, seed1: int, tree_type2, seed2: int
+):
+    """Test the AncestorDescendantSimilarity class of
+    scyphylo against Laura Quintas implementation."""
```
gordonkoehn, line 170: orthogonal checks fail

gordonkoehn: skipping until we know about `scyphylo`

tests/distances/test_scphylo_wrapper.py
```diff
@@ -63,3 +70,164 @@ def test_another_is_different_and_symmetric(similarity, model_tree) -> None:
...
+            TreeNode.convert_to_anytree_node(tree1),
+            TreeNode.convert_to_anytree_node(tree2),
+        )
+        print("\n")
+        print(f"AD_lq: {result2}")
+    except Exception as e:
+        print("\n")
+        tree1.print_topo()
+        tree2.print_topo()
+        raise e
```
gordonkoehn: Laura's never fails :)

src/pyggdrasil/distances/_scphylo_wrapper.py
```diff
@@ -8,7 +10,10 @@
 
 
 class AncestorDescendantSimilarity(interface.TreeSimilarity):
-    """Ancestor-descendant accuracy."""
+    """Ancestor-descendant accuracy.
```
pawel-czyz, line 13: At this stage I'm actually quite convinced that we should just remove this wrapper from our codebase. Would you mind submitting an issue to [scphylo](https://github.com/faridrashidi/scphylo-tools), with the discrepancy you observed? They already have an issue regarding R–F distance, so perhaps we will at some pint move away from using this entirely... (But if MP3 distance is working fine, then we can leave it for now.)

tests/distances/test_scphylo_wrapper.py
pawel-czyz: Nice! Here it's clear what it is for, but generally it's good to have precise docstrings what exactly the returned function is (as `Callable[[arg_types], output_type]` is often not enough to say what function is intended to do and what the arguments mean)

tests/distances/test_scphylo_wrapper.py
pawel-czyz: (Let's leave scphylo's AD, then. Apologies for this issue, I should've checked their test coverage before proposing this package!)

pawel-czyz (approved):
Nice! Apologies for the problems with Ancestor-Descendant metric. I think we can drop `scphylo`'s wrapper entirely. It's good that you investigated it and consulted Laura! :slightly_smiling_face:

tests/distances/test_scphylo_wrapper.py
gordonkoehn: OK understood - as it's just a test I didn't document too well. Thank you.

src/pyggdrasil/distances/_scphylo_wrapper.py
gordonkoehn: Opened an issue here 
- https://github.com/faridrashidi/scphylo-tools/issues/8
and documented as well as I could.

I'll merge this feature as it is now.

PR #103, opened 2023-07-03
Reinstating scphylo Ancestor-Descendant Metric

gordonkoehn:
After a brief exchange with Farid from scphylo I have decided to reinstate his implementation of the AD distance. 
See our helpful exchange here:
- https://github.com/faridrashidi/scphylo-tools/issues/8

It appears that AD is not uniquely defined. Scphylo considers only pairs of mutations, whilst I considered all node pairs including the root node. The literature does not seem explicit about this. See best explanation here https://dl.acm.org/doi/10.1145/3233547.3233584

The scphylo metric throws errors if the first tree is a star tree, i.e. having no ancestor-descendant pairs. Farid, said he'll catch this behaviour. For now, I added a `try catch` and return a `nan` as in those cases no AD exists.

 Laura considers the AD with the root, yet this effectively adds a bias for us by comparing if nodes exist in both trees. 
 As we have all mutations present in our simulated trees, I reinstated the scphylo metric. 
 
 This means that we won't have distances for the AD if the star tree is the initial tree. - Unfortunate, but rather we have different distances to compensate.

### DIFF
--- a/src/pyggdrasil/distances/_scphylo_wrapper.py
+++ b/src/pyggdrasil/distances/_scphylo_wrapper.py
@@ -4,6 +4,7 @@ import warnings
 import scphylo
 
 import anytree
+import jax.numpy as jnp
 
 import pyggdrasil._scphylo_utils as utils
 import pyggdrasil.distances._interface as interface
@@ -12,7 +13,10 @@ import pyggdrasil.distances._interface as interface
 class AncestorDescendantSimilarity(interface.TreeSimilarity):
     """Ancestor-descendant accuracy.
 
-    Note: This function might be damaged. See AncestorDescendantSimilarity_lq instead.
+    Note: - Considers only ancestor-descendant relationships between mutation,
+          i.e. excludes the root node. For an implementation with the root considered
+           see AncestorDescendantSimilarityInclRoot instead.
+          - returns NaN if input tree is star tree.
     """
 
     def calculate(self, /, tree1: anytree.Node, tree2: anytree.Node) -> float:
@@ -32,7 +36,12 @@ class AncestorDescendantSimilarity(interface.TreeSimilarity):
 
         df1 = utils.tree_to_dataframe(tree1)
         df2 = utils.tree_to_dataframe(tree2)
-        return scphylo.tl.ad(df1, df2)
+        try:
+            return scphylo.tl.ad(df1, df2)
+        except ZeroDivisionError:
+            # Known to happen when no pairs of mutations exist in input tree.
+            # i.e., star tree
+            return jnp.nan
 
     def is_symmetric(self) -> bool:
         """Returns ``True`` if the similarity function is symmetric,
--- a/src/pyggdrasil/distances/_scphylo_wrapper.py
+++ b/src/pyggdrasil/distances/_scphylo_wrapper.py
@@ -4,7 +4,6 @@ import warnings
 import scphylo
 
 import anytree
-import jax.numpy as jnp
 
 import pyggdrasil._scphylo_utils as utils
 import pyggdrasil.distances._interface as interface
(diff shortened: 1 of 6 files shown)


src/pyggdrasil/distances/_scphylo_wrapper.py
```diff
@@ -32,7 +36,12 @@ def calculate(self, /, tree1: anytree.Node, tree2: anytree.Node) -> float:
 
         df1 = utils.tree_to_dataframe(tree1)
         df2 = utils.tree_to_dataframe(tree2)
-        return scphylo.tl.ad(df1, df2)
+        try:
```
pawel-czyz, line 39: I guess raising the error is better:

```
try:
    return scphylo.tl.ad(df1, df2)
except ZeroDivisionError:
    raise ZeroDivisionError("Error known to happen when no pairs of mutations exist in input tree. Consider <the name of the improved metric>") 
```

In particular, NaNs could be propagated further down and then e.g., averages or some plots will have some unexpected behaviour.

pawel-czyz (approved):
Very nice work and great that you submitted your very first issue at another repository!

I'd suggest to keep raising the error, so that we don't have troubles with NaN's down in the pipeline.

src/pyggdrasil/distances/_scphylo_wrapper.py
pawel-czyz: Good point! What do you think about something like this?
 
```
class AncestorDescendantSimilarity(interface.TreeSimilarity):
    def __init__(self, raise_error: bool) -> None:
        self._raise_error = raise_error
    
    ...

        try:
            distance = scphylo.tl.ad(df1, df2)
        except ZeroDivisionError:
            if self._raise_error:
               raise ValueError("Error known to happen when no pairs of mutations exist in input tree.")
            else:
                return jnp.nan
```

gordonkoehn: Yes - that surely works. 

Actually, that is just what Farid just did in this [commit](https://github.com/faridrashidi/scphylo-tools/commit/6f3c65cd41c80926eefa700735f85a71bc42ca0c) 

Perhaps we should update our fork  - Or is it easier to just do this ourselves? 

But that still leaves the question of what the snakemake workflow is supposed to do when calculating distances of star trees. The two options seem to be:

1) excluding them from the workflow - i.e. not generating the files.  - implies a somewhat cumbersome output filepath generation for all star tree / AD experiments
2) generate `nan` and empty plots

I am inclined to the first option.

workflows/mark01.smk
```diff
@@ -56,18 +56,21 @@ def make_all_mark01()->list[str]:
...
+                tree_id = TreeId(tree_type=TreeType(tree_type), n_nodes=n_node, seed=tree_seed)
+
+                for n_cell in n_cells:
+                        for error_name, error in errors.items():
+                            for metric in metrics:
+                                # AD is not defined for star trees - skip this case
+                                if tree_type == "s" and metric == "AD":
+                                    continue
+                                filepaths.append(filepath+f"{tree_id}-{n_cell}_{error['fpr']}_{error['fnr']}_0.0_{observe_homozygous}_{cell_attachment_strategy}/{metric}_hist.svg")
+
```
gordonkoehn, line 73: To avoid a failing snakemake workflow I have not skipped all special cases of the initial star tree and AD metric. 

Can do this from now on - just means one think to take care of when using the Ad.

src/pyggdrasil/distances/_scphylo_wrapper.py
pawel-czyz: As I haven't touched the fork for quite some time, I worry that it may have diverged by quite a lot and trying to update it would lead to all sorts of compatibility issues :sweat_smile: 

As your solution below works we can drop modifying our A-D distance as well :slightly_smiling_face:

pawel-czyz (approved):
I think this works for us, thanks for investigating the issue!

Issue #108, opened 2023-07-05
Migrate Tree Inference Snakemake Rules from scripts to run directive 


gordonkoehn:
Remove "REPODIR" in `tree_inference.smk`, by removing scripts, i.e. by run directives in snakemake rules.

Currently the directory of the repository root of `PYggdrasil` has to be entered in the `tree_inference.smk`, because the snakemake rules need global paths to the scripts. 

This somewhat cumbersome dependency can be removed by migrating all of the logic from the scripts into run directives.

May also allow for better type-checking.

PR #116, opened 2023-07-21
Removing tree script dependence of workflows

gordonkoehn:
The current implementation of the snakemake workflows requires setting two file paths. Once in the experiment to run `markXX.smk` and once in the `tree_inference.smk`, that is because some of the base rules in tree inference run scripts instead of bare python. 

This dependence on a filepath is cumbersome and error-prone. It is best to remove it for usability and to remove the scripts altogether. Therefore this PR will integrate the functionality of the scripts into the package and rewrite the snakemake rules to pure Python. 

- [x] tree generation and `make_tree.py`

This is one of three PRs to address this issue
- #108


workflows/mark00.smk
```diff
@@ -4,8 +4,8 @@ from pyggdrasil.tree_inference import McmcConfig, TreeId, TreeType, CellSimulati
 
 ################################################################################
 # Environment variables
-DATADIR = "/cluster/work/bewi/members/gkoehn/data"
-#WORKDIR = "../data"
+#DATADIR = "/cluster/work/bewi/members/gkoehn/data"
+DATADIR = "../data"
```
gordonkoehn, line 8: Currently - I have to change this every time I run it on a different machine - What's the best way to do this ?

Actually, have a config file here I suppose.

PR #125, opened 2023-07-25
Fixing Mark03 rule error

gordonkoehn:
the log_prob plots were missing inputs. 

Needed to add another dirty rule to make this work.

This PR also implements some fixed for MARK02. 
The initial trees for MAKR02 were the true trees twice.

### DIFF
--- a/workflows/mark02.smk
+++ b/workflows/mark02.smk
@@ -56,7 +56,7 @@ CS_seed = 42 # <-- configure cell simulation seed here
 #####################
 # True Tree Parameters
 tree_types = ["r"] # <-- configure tree type here ["r","s","d"]
-tree_seeds = [42, 34] # <-- configure tree seed here
+tree_seeds = [42]#, 34] # <-- configure tree seed here
 
 #####################
 #####################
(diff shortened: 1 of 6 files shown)


workflows/mark02.smk
```diff
@@ -66,8 +66,8 @@ tree_seeds = [42, 34] # <-- configure tree seed here
 # given each error rate, true tree, no of cells and mutations
 initial_points = [ # (mcmc_seed, init_tree_type, init_tree_seed)
     (42, 'r', 45),
-    (12, 'r', 34),
-    (34, 'r', 42),
+    (12, 'r', 20),
+    (34, 'r', 31),
```
gordonkoehn, line 70: See Initial Trees For Mark02 were changed here.

workflows/mark03.smk
```diff
@@ -704,4 +704,39 @@ rule combined_logProb_iteration_plot_noHuntress:
...
+        log_probs = mcmc_data.log_probabilities
+        # convert log probs to list of float
+        log_probs = [float(i) for i in log_probs]
+        # get iteration
+        iteration = mcmc_data.iterations
+        # convert iterations of list of int
+        iteration = [int(i) for i in iteration]
+
+        # save
+        yg.serialize.save_metric_result(iteration,log_probs,fp)
```
gordonkoehn, line 742: Dirty Rule needed for MARK03 - To save the log_prob true tree is needed in the path even though the calculation of log_prib does not depend on it.

Issue #127, opened 2023-07-26
Switch/Update to other HUNTRESS Implementation


gordonkoehn:
It has been discovered that the `scphylo` implementation does fail silently when running it with multiple threads as essential for larger cell-mutation profiles.

This has been confirmed by the developers.
- https://github.com/faridrashidi/scphylo-tools/issues/10#issuecomment-1650778632

The issue has been addressed by the developers - and we shall update our scphylo dependency. 

- [ ] check why @pawel-czyz did a fork to start with / Why did we remove the `numba` dependency?

gordonkoehn:
I'll try to just update the package for now.

gordonkoehn:
Just updating to the newest version of `scyphlo works` when downgrading `nextworkX`, but I get trouble with `numpy` and `numba` - I am not sure why, something about a `long` not being defined. 

Perhaps passing variables back to my numpy environment brings trouble - really don't know. Let's try to do a fork like @pawel-czyz  did on my own, removing `netwrokX` and `numba` again.

gordonkoehn:
Make my own fork removing `mamba` and unpining the `networkX` version. All tests work.
- https://github.com/gordonkoehn/scphylo-tools

gordonkoehn:
HUNTRESS indeed seems to work for multiple threads here. 
- [ ] Test 8 cores 16 GB ram locally for 1000 cells and 10 mutations.

gordonkoehn:
Merge before testing. - Try in HPC EULER.

PR #128, opened 2023-07-26
Update scphylo / fix  HUNTRESS implementation

gordonkoehn:
See
- #127


gordonkoehn:
@pawel-czyz FYI, as `scphylo`'s HUNTRESS was updated and I hoped it would fix the multithreading issued I switched from your fork to one I made.

pawel-czyz:
Good idea, thanks for handling this! :slightly_smiling_face:

PR #140, opened 2023-08-04
Fix of Range of Combined MARK01 plots

gordonkoehn:
The histogram range seems broken on the plots generated so far.

### DIFF
--- a/src/pyggdrasil/serialize/_to_json.py
+++ b/src/pyggdrasil/serialize/_to_json.py
@@ -220,4 +220,7 @@ def read_metric_result(fullpath: Path) -> tuple[list[Union[int, str]], list[floa
     # get axis name
     axis_name = list(metric_dict.keys())[0]
 
+    # convert metric result to list of floats
+    metric_dict["result"] = [float(x) for x in metric_dict["result"]]
+
     return metric_dict[axis_name], metric_dict["result"]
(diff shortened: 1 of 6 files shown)


gordonkoehn:
combined_AD_hist
combined_DL_hist
combined_MP3_hist

src/pyggdrasil/serialize/_to_json.py
```diff
@@ -220,4 +220,7 @@ def read_metric_result(fullpath: Path) -> tuple[list[Union[int, str]], list[floa
     # get axis name
     axis_name = list(metric_dict.keys())[0]
 
+    # convert metric result to list of floats
+    metric_dict["result"] = [float(x) for x in metric_dict["result"]]
+
```
gordonkoehn, line 225: just making sure these are indeed floats

workflows/mark01.smk
```diff
@@ -150,15 +150,16 @@ rule make_combined_histograms:
...
             hist_data = distances[i, 1, :]  # Select the 2nd position data for the i-th element
             error_name = list(errors.keys())[i]
-            plot_label = plot_label + [error_name]
-            plot_data = plot_data + [hist_data]
-        axs.hist(plot_data,bins='auto', range=[0,1],color=colors,label=plot_label)
+            plot_label.append(error_name)
+            plot_data.append(np.array(hist_data).astype(float))
+        # plot the histogram
+        axs.hist(plot_data,range=(0,1),color=colors,label=plot_label, histtype='bar')
         # Put a legend to the right of the current axis
```
gordonkoehn, line 157: Just issues with floats and str.

PR #148, opened 2023-08-11
Same Likelihood Investigation

gordonkoehn:
**NOT MEANT TO ME MERGED - INVESTIGATION ONLY**

I am looking at the HUNTRESS case for this plot of Experiment mark03. 

The log-prob does not seem to change EVER here. 

Yet there are 60 different trees if I am not mistaken.

Screenshot 2023-08-11 at 15 52 44

This seems very problematic. 

I would really like to verify the log-prob calculation.


gordonkoehn:
@pawel-czyz Here is the investigation you wanted me to do. 

We already looked at the mutation data for this case and it looked good. 

This is a big issue or big bug. ;/

gordonkoehn:
will add more investigations on 30 min ... not completely uploaded

gordonkoehn:
@pawel-czyz see the full notebook here

https://github.com/cbg-ethz/PYggdrasil/blob/gordon/bassin_investigation/notebooks/basin_likelihood.ipynb

pawel-czyz:
Very nice, thanks! I think another implementation of the log-probability function can help resolve the issue whether it's a bug or not: implementing log-probability efficiently is was hard task, as there are many possible places where a hard-to-avoid bug could enter the code (e.g., there are risks of off-by-one error or typical problems with keeping orders of dimensions in arrays consistent). 

I'd suggest to implement a slow, but very simple and interpretable one, log-probability function.
For example, without using `einsums` over multi-dimensional arrays, but calculating $\log P(\mathrm{cell} | \mathrm{tree})$ individually for each cell, based on simple expression (e.g., each gene separately, rather than `jnp.where`).

gordonkoehn:
@pawel-czyz 
Eureka! - I found the issue. - I know how to solve the `log-prob` - 80 % certain.
 Something completely different than expected. 

I'll walk you through it tomorrow - let's discuss how to fix it then. Put it here for documentation.

___

The `log_prob` implementations are all fine. But they assume that the tree in its `Tree` format is ordered, i.e. it's labels are going from `0,1,2, n+1`.

That assumption is always satisfied for tests, as that is a natural way to propose a tree. 

During the MCMC moves the order of nodes gets scrambled. That is no problem with the way the tree is stored 'Tree', nor to modify it further. The trees are also plotted with some effort of reordering using the `Anytree / TreeNode` framework, but the core of it all the `log_probability` does assume ordered node labels.

Even the orthogonal implementation of the `logprob_validator.logprobability_fn` would have failed partially as it still assumes to root to come last. 

The reason the tree moves scramble the order is because it seemed just less error-prone to implement, with adjacency matrices. I spend a long time to get those MCMC moves right. The usage of multiple tree formats has certainly contributed to some opacity here.

The easy way to fix this is 
1) to reorder the tree node labels and adjacency matrices after each tree move
2) Alternatively one has to iterate through the explicit node labels, no tensor implementation there I can imagine.
3) redo the MCMC moves to not scramble the order - had a lot of issues with this - not recommended!

The functionality to implement the reordering already exist from conversions to `TreeNodes`. 

This sounded plausible but one mystery remained. How can we end up getting 60 different Trees of exactly the same `log_prob`. 

Screenshot 2023-08-22 at 20 47 21

Notice, here that the trees plotted are indeed different and their adjacency matrix representation is also different by the order of tree labels. BUT the adjacency matrix is the SAME! I went through all the log file and it is the same for all the `Tree` representations. 
That seemed very odd - how can these matrices all look the same but still represent the same trees. And wouldn't a swap subtrees move definitely change the adjacency matrix?

I considered this, but assuming the trees, only ever have two levels, i.e. a cell has at most two mutations plus root - trees are no deeper than two levels. --> all adjacency matrices look the same, only the label order changes.

This may just be bad luck - deep trees don't come around often.

Now 60 trees make perfect sense. As only 3 nodes have to be ordered for two-level trees, `5x4x3 = 60`

This is the cruellest bug I have produced so far in my life!

pawel-czyz:
I see, good you found it out!

How about reordering? This can be accomplished in JAX by building a permutation matrix $P$ (initializing with zeros and then iterating over the labels vector) and then using a formula like
$$A' = P^TAP$$
which should be very fast in JAX.

PR #149, opened 2023-08-14
Verify / Add Tests of Log-Probability Calculation

gordonkoehn:
Verifying that the log probability is calculated correctly.

The calculation is split into the calculation of

- [x] mutation likelihood 
$$P(D_{ij} | A(T)_{ik})$$
- [x] log mutation likelihood 
$$\log P(D_{ij} | A(T)_{ik})$$
- [x] log-probability 
 $$\sum_{j=1}^m \log \sum_{k=1}^{n+1} \exp \sum_{j=1}^{n} \log P(D_{ij} | A(T)_{ik})$$


gordonkoehn:
I went through the mutation likelihood calculations - there are examples and exact by-hand calculations for unit tests. This is likely to be correct.

gordonkoehn:
log mutation likelihood is also just element-wise calculation of the logarithm - nothing error prone here

gordonkoehn:
Do a manual calculation of log prob with two cells 3 mutations on paper.

gordonkoehn:
I manually calculated the log-probability for an error-free mutation matrix, with 2 cells and 3 mutations.

The result agrees with the one by our current implementation up to at least 10 decimals.

[ManualLogProb.pdf](https://github.com/cbg-ethz/PYggdrasil/files/12344639/ManualLogProb.pdf)

tests/tree_inference/test_logprob.py
```diff
@@ -192,3 +192,36 @@ def test_logprobability_fn_direction():
...
+    beta = 0.3
+    theta = (alpha, beta)
+    # should result in a (n)* (n+1) * (m) tensor or all elements 0.5
+
+    # expected
+    # manually compute expected logprob
+    # expected = -0.79861
+    expected = jnp.log(
+        0.3 * 0.7 * 0.9 + 0.9 * 0.7 * 0.9 + 0.9 * 0.1 * 0.3 + 0.9 * 0.1 * 0.9
+    ) + jnp.log(0.7 * 0.7 * 0.9 + 0.1 * 0.7 * 0.9 + 0.1 * 0.1 * 0.3 + 0.1 * 0.1 * 0.9)
```
gordonkoehn, line 223: Handheld calculator and jnp.log jax gives slightly different values

pawel-czyz:
Another validation technique came to my mind: for the data set with 5 mutations 200+ cells, small noise and two trees differing in 0-2 edge (or another pair of two trees with minor differences) giving the same total loglikelihood.

What are the loglikelihoods of *individual cells*? Perhaps this can pinpoint if there's an issue...

tests/tree_inference/test_logprob.py
```diff
@@ -192,3 +192,94 @@ def test_logprobability_fn_direction():
...
+    log_prob = logprob.logprobability_fn(mutation_matrix, tree, theta)
+    assert jnp.isclose(log_prob, expected, atol=1e-10)
+
+
+def test_mutation_likelihood_fn_exact_m2n3():
+    """Test mutation_likelihood_fn manually.
+
+    Manual mutation likelihood calculation for a tree with 2 mutations and 3 cells.
+    See https://github.com/cbg-ethz/PYggdrasil/pull/149 for calculation.
+    """
```
gordonkoehn, line 234: Added another manual test of the mutation likelihood.

src/pyggdrasil/tree_inference/_logprob.py
```diff
@@ -171,3 +171,94 @@ def _mutation_likelihood(
...
+    """Calculates the log-probability of a tree given error rates and data.
+        Uses basic for loops to verify the vectorized implementation.
+    Args:
+        data: observed mutation matrix to calculate the log-probability of
+        tree: tree to calculate the log-probability of
+        theta: \theta = (\fpr, \fnr) error rates
+    Returns:
+        log-probability of the tree
+        Implements the log-probability function from the paper: Equation 13
+        P(D|T,\theta) = \frac{T,\theta | D}{P(T,\theta)}
```
gordonkoehn: This function verified that given the `log_mutation_likelihood` the application of einsum and log sum exp tricks is done correctly.

gordonkoehn:
Added two implementations of the log_probability now - both are different and written independently.

Unfortunately for now the number of solutions scales with the number of implementations. 

I'll have to debug both next.

gordonkoehn:
Also design a test that compares if the noisy mutation matrix gets a lower score than the perfect one for all three implementations.

gordonkoehn:
Also one can for sure calculate method 3 my hand.

gordonkoehn:
- [ ] devise a manual test to check the validator of perfect data for a true tree and a wrong contradicting tree, that should be resolved by the cell attachment implied by the perfect data. 
- [ ] do a manual calculation of the lob prob as done in the validator

gordonkoehn:
I implemented a test for both lob-probs. The test checks if the true tree is better or equal in log-prob compared to a tree that is different and modified by one MCMC move under SCITE. The data is perfect, no noise and we have 10x more cells than mutations, so the tree structure should be resolvable.

Surprisingly, both log-prob functions pass this test. I can it for quite a few small trees.
 
 Fast log-prob
Screenshot 2023-08-21 at 14 35 55

Slow-log-prob / validator
Screenshot 2023-08-21 at 14 37 34

Both functions are able to distinguish the trees, but the scales are different. 

This is mysterious. Perhaps an indication that some exponentiation is not done in the fast implementation? 

I'll calculate a small tree by hand now.

@pawel-czyz What are your thoughts?

pawel-czyz:
> Surprisingly, both log-prob functions pass this test. I can it for quite a few small trees.

I like this test, it's a good sanity check! However, I think it can have large false negative rate (i.e., many functions may have this property).

The scales are indeed worrying. There are two things which may help resolve the issue:

  - An implementation in terms of very, very small functions, so that everything is well-tested at each phase. (I.e., `(true tree, attachment_index) -> true mutations vector of one cell` as one function `(true mutations vector, rates, noisy mutations vector -> loglikelihood`  and marginalisation over attachments all in different functions).
  - Perhaps it'd be good to bring the normalization terms in the loglikelihood (that is, the probability of cell attachment is $1/n_\text{nodes}$, rather than $1$). Then, $\sum_\text{data} P(\text{data} \mid \text{tree}, \text{rates} ) =1$. For e.g., 1 cells and 10 mutations it's $2^{10}$ options for the data vector, so it should be tractable to enumerate over all of them and check if the sum is right.

gordonkoehn:
@pawel-czyz 

This investigation was a rollercoaster. Here is the summary. I implemented two partially and completely orthogonal versions of the `log_probability_fn`, the latter uses only `loops` and `vectors`. 

I have validated it all by hand for small trees and am 99% certain that the SLOW orthogonal implementation does how I understand SCITE works. - Put it this way, I am less certain I understand SCITE correctly.

Besides the orthogonal test between all three implementations, I have implemented tests for:
1) Do we score a true tree higher, given it is identifiable (many cells, no noise), than a tree one MCMC move away?
2) Do we score a tree higher under the perfect mutation matrix compared to the noisy one.  

For small trees and cell numbers, the world is beautiful. - all tests check out.

I have only run the orthogonal test for many cells so far and the picture is confusing to me. I check their outputs to `10^-6` tolerance. Is that to rigorous for jax / numpy / base python. Have no good foundation for this tolerance value. What shall we allow?

For few cells and mutations all implementations agree. For 20 mutations and 200 cells, they don't. Their values aren't far off. 

I am uncertain if this is just the normal trouble with computers or if this may be the reason for the unidentifiablility problem we observed for 200 cells and 5 mutations.

I'll investigate.

src/pyggdrasil/tree_inference/_logprob_validator.py
```diff
@@ -0,0 +1,173 @@
...
+Implements a dumb version of the log-probability function, which is used for testing.
+
+This version attempts to do it by intuition without looking ta the paper.
+"""
+import jax
+import jax.numpy as jnp
+
+import pyggdrasil.tree_inference._tree as tr
+
+from pyggdrasil.tree_inference import Tree, ErrorRates
```
pawel-czyz, line 14: This is a place where a potential circular import could appear. One way around this would be to import these from `_interface` submodule

pawel-czyz:
> This investigation was a rollercoaster. Here is the summary. I implemented two partially and completely orthogonal versions of the log_probability_fn, the latter uses only loops and vectors.

Cool!

> I have validated it all by hand for small trees and am 99% certain that the SLOW orthogonal implementation does how I understand SCITE works. - Put it this way, I am less certain I understand SCITE correctly.

Let's discuss it with a whiteboard, then :slightly_smiling_face: 

> Besides the orthogonal test between all three implementations, I have implemented tests for:
>
> Do we score a true tree higher, given it is identifiable (many cells, no noise), than a tree one MCMC move away?
> Do we score a tree higher under the perfect mutation matrix compared to the noisy one.
> For small trees and cell numbers, the world is beautiful. - all tests check out.

Great!

> I have only run the orthogonal test for many cells so far and the picture is confusing to me. I check their outputs to 10^-6 tolerance. Is that to rigorous for jax / numpy / base python. Have no good foundation for this tolerance value.
> What shall we allow?

*Log-probability* up to $10^{-6}$ is definitely good enough to me! (The *probability* not really – probabilities of most trees should be close to 0.)

> For few cells and mutations all implementations agree. For 20 mutations and 200 cells, they don't. Their values aren't far off.
>
> I am uncertain if this is just the normal trouble with computers or if this may be the reason for the unidentifiablility problem we observed for 200 cells and 5 mutations.
> I'll investigate.

I 've added some comments on how to make it more numerically stable :slightly_smiling_face:

tests/tree_inference/test_logprob.py
```diff
@@ -192,3 +194,213 @@ def test_logprobability_fn_direction():
...
+
+
+@pytest.mark.parametrize("n_cells", [40])
+@pytest.mark.parametrize("n_mutations", [2, 4])
+@pytest.mark.parametrize("seed", [23, 2, 5])
+def test_logprob_no_noise_many_cells_wrong_tree(
+    n_cells: int, n_mutations: int, seed: int
+):
+    """Test if given no noise and many cell small trees can be identified.
+    make a small true tree, and compare its likelihood to that of a mcmc 1 tree."""
```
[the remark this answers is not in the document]
pawel-czyz: Do these 60 trees also have the same log-probability under the new implementation?

gordonkoehn:
-[ ] Add test to check for different log probability after swapping nodes ?

src/pyggdrasil/tree_inference/_logprob_validator.py
```diff
@@ -0,0 +1,173 @@
...
+        tree: tree to calculate the expected likelihood of
+        cell_attachment: cell attachment index (node index attached to),
+            for a single cell
+        mutation_i : mutation index
+
+    Returns:
+        expected likelihood of the tree
+    """
+    # get the ancestor matrix
+    ancestor_mat = tr._get_ancestor_matrix(tree.tree_topology)
```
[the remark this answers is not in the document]
gordonkoehn: Yes, indeed! This is what you drew on paper. Getting a list of parent nodes of node x, excluding the root, including the node itself. Then check if the node/mutaiton is in that list given that we attach a cell to node x.

src/pyggdrasil/tree_inference/_logprob_validator.py
```diff
@@ -0,0 +1,173 @@
...
+    log_prob = jnp.log(p_cell_mutation_attachment)
+    logger.debug(f"log_prob={log_prob}")
+
+    return float(log_prob)
+
+
+def _exp_sum_mutations(
+    tree: Tree, data: jax.Array, error_rates: ErrorRates, cell_attachment: int
+) -> float:
+    """Returns the exponential sum of the log-probabilities of all
```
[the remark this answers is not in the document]
gordonkoehn: This is great. - Thank you for sharing this with me! 

Really enjoy these little readings.

src/pyggdrasil/tree_inference/_logprob_validator.py
```diff
@@ -0,0 +1,173 @@
...
+        mutations for a given cell and attachment.
+
+    Args:
+        data is a column of the data matrix, for a given cell
+
+    """
+    # all but the last label, last is root
+    mutations = tree.labels[:-1]
+    sum = 0
```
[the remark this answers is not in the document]
gordonkoehn: This is great. - Thank you for sharing this with me!

Really enjoy these little readings.

src/pyggdrasil/tree_inference/_logprob_validator.py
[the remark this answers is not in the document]
gordonkoehn: Here we have the the latter

src/pyggdrasil/tree_inference/_logprob_validator.py
gordonkoehn: Fixed!

src/pyggdrasil/tree_inference/_logprob.py
```diff
@@ -50,12 +52,14 @@ def logprob_(tree) -> float:
     return logprob_
 
 
-def logprobability_fn(data: MutationMatrix, tree: tr.Tree, theta: ErrorRates) -> float:
+def logprobability_fn(
+    data: MutationMatrix, tree: OrderedTree, theta: ErrorRates
+) -> float:
```
gordonkoehn, line 57: Now I am typing to use `OrderedTree`s in the `log_probability_fn`.

src/pyggdrasil/tree_inference/_mcmc.py
```diff
@@ -104,7 +109,11 @@ def _prune_and_reattach_proposal(key: JAXRandomKey, tree: Tree) -> Tuple[Tree, f
 
 def _swap_node_labels_move(tree: Tree, node1: int, node2: int) -> Tree:
     """Swaps labels between ``node1`` and ``node2`` leaving the tree topology
-    untouched."""
+    untouched.
+
+    Note:
+        - Changes the labels of the tree, requires a reordering of the tree.
+    """
```
gordonkoehn, line 116: The culprit was here !! :)

src/pyggdrasil/tree_inference/_mcmc.py
```diff
@@ -402,6 +411,9 @@ def _mcmc_kernel(
     else:
         proposal, log_q_diff = _swap_subtrees_proposal(key, tree)
 
+    # reorder tree, for 0-N indexing for log-probability function
+    proposal = OrderedTree.from_tree(proposal)
+
```
gordonkoehn, line 416: The FIX! - Regardless of the move I reorder the proposed tree in the MCMC kernel now.

src/pyggdrasil/tree_inference/_ordered_tree.py
```diff
@@ -0,0 +1,63 @@
...
+        Args:
+            tree: Tree to convert to OrderedTree
+
+        Returns:
+            OrderedTree
+        """
+        # reorder tree
+        ordered_tree = _reorder_tree(tree, jnp.arange(tree.labels.shape[0]))
+        # return ordered tree type
+        return OrderedTree(ordered_tree.tree_topology, ordered_tree.labels)
```
gordonkoehn, line 63: Created a class to clarify where `OrderedTree`s are required, i.e. the MCMC kernel. 

This class `assert`s every tree ever instantiated. 

Might be a little paranoid - What do you think?

src/pyggdrasil/tree_inference/_tree.py
```diff
@@ -267,9 +268,11 @@ def _get_root_label(tree: Tree) -> int:
     return root_label
 
 
-def _reorder_tree(tree: Tree, from_labels, to_labels):
+def _reorder_tree_verify(tree: Tree, to_labels):
     """Reorders tree from current labels to new labels
 
+    Uses loop to reorder tree - SLOW. >100x slower than _reorder_tree.
+
```
gordonkoehn, line 275: Legacy / Orthogonal implementation of `_reorder_tree`. - very slow.

tests/tree_inference/test_logprob.py
```diff
@@ -161,7 +163,7 @@ def test_logprobability_fn():
     expected = 5 * jnp.log(4 * jnp.exp(3 * jnp.log(0.5)))
 
     # define logprob fn
-    log_prob = logprob.logprobability_fn(mutation_matrix, tree, theta)
+    log_prob = logprob.logprobability_fn(mutation_matrix, tree, theta)  # type: ignore
```
gordonkoehn, line 166: Note all tests of `log_prob` functions ignore the types of trees inputted now as a consequence of enforcing the `OrderedTree`s

tests/tree_inference/test_tree.py
```diff
@@ -254,3 +254,31 @@ def test_is_valid_tree(n: int, seed: int, tree_type: str):
...
     tree = make_tree(n, seed, tree_type)
     assert tr.is_valid_tree(tree) is True
+
+
+@pytest.mark.parametrize("n", [3, 4])
+@pytest.mark.parametrize("seed", [424, 32])
+@pytest.mark.parametrize("tree_type", ["r", "d", "s"])
+def test_verify_reorder_trees(n: int, seed: int, tree_type: str):
+    """Tests the orthogonal implementations of reordering trees."""
+
```
gordonkoehn, line 264: Orthogonal test of `_reorder` functions.

gordonkoehn (commented):
Continuing from the `log_prob` tests, I have implemented the tree reordering to fix the issue with the calculation of the fast `log_prob`. 

Now proposed trees will be reordered to have `0,...,N` labels, compatible with the fast calculation of the `log_probability_fn`. For that matter I have implemented a fast `_reorder_tree` function based on a permutation matrix operation. 

To further clarify the usage of ordered trees I have added a subclass of `Tree` that asserts the labels called `OrderedTree`. 

Please let me know what you think :) 

Next I'll get onto the MARK04 experiment.... performance of HUNTRESS with many trees.

src/pyggdrasil/tree_inference/_ordered_tree.py
pawel-czyz: I think it's good – we can always optimize if needed (probably not needed), using this as a reference implementation :slightly_smiling_face:

pawel-czyz (approved):
Well-done! :partying_face: Wonderful investigation and the fix :slightly_smiling_face: 

I wonder: how many posterior modes seem to be now? Would it be possible to re-run the experiments?

src/pyggdrasil/tree_inference/_mcmc.py
gordonkoehn: Yes - they should as tree generation always gives ordered trees. 

Now that I consider this again, it might be violated by HUNTRESS. 

I added a reordering in the MCMC sampler. 

Well spotted!

src/pyggdrasil/tree_inference/_tree.py
```diff
@@ -296,6 +300,31 @@ def _reorder_tree(tree: Tree, from_labels, to_labels):
...
+    adj_matrix = tree.tree_topology
+    labels = tree.labels
+    # Create a permutation matrix based on the mapping
+    P = jnp.zeros((len(labels), len(labels)), dtype=jnp.int32)
+    matching_indices = np.where(labels[:, None] == new_labels)
+    P = P.at[matching_indices[0], matching_indices[1]].set(1)
+    # Reorder the adjacency matrix using matrix multiplication
+    reordered_adj_matrix = P.T @ adj_matrix @ P
+    # make and return tree
+    return Tree(reordered_adj_matrix, new_labels)
```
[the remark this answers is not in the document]
gordonkoehn: Thanks for the suggestion. I added a shape argument to the `jnp.where` hoping it helps jax. 

I fail to see how I could use `jnp.eye` here. Could you please help me see how ? 
@pawel-czyz

Issue #145, opened 2023-08-10
Meeting notes from August 9th


pawel-czyz:
This is not an issue per se, but rather a summary of the feedback and idea we received yesterday from Niko, Jack, Joanna, Johannes and Xiang, so we can decide on the future directions.

## Distances and tree summary statistics
  - MP3 seems to be too noisy for our purposes.
  - We need to decide on good distance to use. Perfectly, interpretable and which makes sense for downstream applications as:
    - Is the order of mutations $A$ and $B$ preserved? (Some variations on the A–D distance).
    - Are the pathways preserved? (Some variations on DL distance; with applications for TreeMHN in mind).
    - For developmental biology and cell clustering: how many clones exist? What are the lineages and how closely different cells are?
  - We considered also other summary statistic, as lengths of the paths (e.g., tree depth or median), node degrees.
  - Validate behaviour of the distances by studying the distribution of it when the tree is modified. For example, let $k$ be some transition kernel (e.g., making some kind of move in the tree space with probability $1$). How does the distance distribution $d( \mathcal T, d^{(n)}(\mathcal T) )$ look like for different trees $\mathcal T$ and $n$, where $d^{(n)}(\mathcal T) = d(d(\dotsc d(\mathcal T)))$? How does it look like for "random" trees.

- False positive rate 0.1 is much too large. Reasonable "large" rates include $10^{-4}$, $10^{-5}$, or, for whole exome sequencing, $10^{-2}$.

- Show summaries over several different tree topologies, rather than one fixed tree, to understand the variation caused also by tree selection, rather than just cell sampling.
- Consider better ways of generating the trees and cells:
  - We can perhaps use Robert Noble's tree topologies as examples.
  - We can enforce some clonal mutations.
  - We can generate deeper trees by preferential attachment of nodes to trajectories/Chinese restaurant process/Pitman-Yor process.
  - We can attach cells only to leaves.
  - Or we can have "attachability" which is a function of depth. (The deeper the more likely that a cell attached there would survive).

gordonkoehn:
**Additions:**
- Investigate the potential positive bias of HUNTRESS, with trees and mutation matrices in the same order, and consider scrambling the order of mutations in the mutation matrix w.r.t. the tree - concern voiced by Jack
- Remove claim in repo README that PYggdrasil can be used for copy number inference

These are quite a lot of ideas on what to enhance and investigate. - With SMC this becomes quite the strategic choice what we want to do in the remaining time.

[...]

PR #150, opened 2023-08-15
More Realistic Error Rates

gordonkoehn:
Upon the feedback in the mid-term presentation
- #145 
I added another error rate condition called "Moderate" with a better ration of FPR / FNR. 

Screenshot 2023-08-15 at 16 04 07

**Is this a good set of conditions ?**

Further, I allow to select for error rate conditions in the snakemake experiments now.

### DIFF
--- a/workflows/mark01.smk
+++ b/workflows/mark01.smk
@@ -25,9 +25,14 @@ metrics = ["MP3","AD","DL"]  # <-- configure distances here
 num_samples = 30 # <-- configure number of samples here
 
 # Errors <--- set the error rates here
-errors = {
+selected_error_cond = ['IDEAL', 'TYPICAL','MODERATE', 'LARGE']
+all_error_cond = {
         member.name: member.value.dict()
         for member in yg.tree_inference.ErrorCombinations
+}
+errors = {
+    name: all_error_cond[name]
+    for name in selected_error_cond
 }
 n_mutations = [30] #5, 10, 50] # <-- configure number of mutations here
 n_cells = [200, 1000] #, 5000] # <-- configure number of cells here
--- a/workflows/mark01.smk
+++ b/workflows/mark01.smk
@@ -25,7 +25,7 @@ metrics = ["MP3","AD","DL"]  # <-- configure distances here
 num_samples = 30 # <-- configure number of samples here
 
 # Errors <--- set the error rates here
-selected_error_cond = ['IDEAL', 'TYPICAL','MODERATE', 'LARGE']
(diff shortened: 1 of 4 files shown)


pawel-czyz:
I recall that the extreme FPR is 1%. Perhaps we can do three realistic settings:
```
Ideal: FPR=1e-6, FNR=1e-6
Typical: FPR=1e-6, FNR=0.1
Large: FPR=0.01, FNR=0.2
```
and for selected experiments (but not all) the `Very Very Extreme` (which is not realistic), with e.g., `FPR=FNR=0.45` to just see how quickly the MCMC explores the space in the limit when likelihood is almost non-informative.

workflows/mark01.smk
```diff
@@ -25,10 +25,15 @@ metrics = ["MP3","AD","DL"]  # <-- configure distances here
 num_samples = 30 # <-- configure number of samples here
 
 # Errors <--- set the error rates here
-errors = {
+selected_error_cond = ['IDEAL', 'TYPICAL','MODERATE', 'LARGE']
```
pawel-czyz, line 28: I like this parametrization of rules very much!

gordonkoehn:
I like having fewer more realistic error combinations! :) So let's go for the 4 you propose, with one adjustment - if you agree.

where do you have the 1% FRP from - can't find a reference for that.
 
I went through all references from the mid-term presentation and chats with Johannes and believe we should use the hardest realistic setting as FPR=1e-5 / 1e-4, still 2-3 orders of magnitude over the median somatic mutation rate:

`
Ideal: FPR=1e-6, FNR=1e-6
Typical: FPR=1e-6, FNR=0.1
Large: FPR=1e-4, FNR=0.2

Extreme: FPR=0.45, FPR=0.45
`

Here is what Johannes told me a while back:

>  - Die False-Discovery Rates wurden etwas systematischer in folgendem Paper studiert:
https://www.sciencedirect.com/science/article/pii/S0092867412002280?via%3Dihub#mmc1 (Document S1 ist glaube ich aufschlussreich). Die niedrigste ADO rate die man so erwartet liegt bei etwa 7 %.

>  - In meinen (relativ modernen) Daten liegt die mittlere ADO rate abhängig vom sample bei ca. 12-16%

> - Die False-Positive rate liegt etwa in der Range von 10**-6 bis 10 **-5 und die ADO rate wird auf mehr als 25 % geschätzt in den extremsten fällen (das findest du hier: https://www.nature.com/articles/s41467-021-26938-w#Fig2 in introduction und Figure 2)

gordonkoehn:
ok, will implement as u suggested then.

PR #155, opened 2023-08-22
Fix improper random key usage in MCMC kernel

gordonkoehn:
Double-checking all core components of the MCMC kernel I found an improper, yet retrospectively unproblematic usage of a random key. This small PR fixes this.


pawel-czyz (approved):
Good catch, thanks!

PR #156, opened 2023-08-28
Consider Blocking HUNTRESS Easy Data 

gordonkoehn:
**ON HOLD - NOT TO BE MERGED**

I recall Jack asked if we scramble the order of mutation matrices before we feed them to HUNTRESS. His doubt was that perhaps the order of tree labels and the mutation matrix is somewhat beneficial to the inference performed by HUNTRESS. 

To block this technical artefact we could reorder the mutation matrix rows.

We currently don't. In this PR I considered this quickly. 

The issue is that HUNTRESS does take the mutation matrix indices as the node names. Thus one would have to relabel the nodes again after Huntress is performed. 

This brings a few ways to introduce errors I suppose. 

It is probably best to wait and see if the advantage of HUNTRESS is still so pronounced now that SCITE should be fixed soon.

### DIFF
--- a/src/pyggdrasil/tree_inference/__init__.py
+++ b/src/pyggdrasil/tree_inference/__init__.py
@@ -41,6 +41,7 @@ from pyggdrasil.tree_inference._simulate import (
     CellSimulationModel,
     get_simulation_data,
     CellSimulationData,
+    scramble_node_order_mutations
 )
 
 from pyggdrasil.tree_inference._file_id import (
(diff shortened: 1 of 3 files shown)


gordonkoehn:
The effect of the initial tree fades quickly. Thus, the usage of HUNTRESS appears to be justified, despite its great performance. This investigation shall be closed as it no longer seems worth the pursuit.

PR #157, opened 2023-08-28
Implementing MARK04 - How well does HUNTRESS work for different trees ?

gordonkoehn:
This PR implements another HUNTRESS experiment, that is to investigate how tree-dependent the performance of HUNTRESS is.

This experiment is a variation of MARK01 - code largely copied.

### DIFF
--- a/workflows/Snakefile
+++ b/workflows/Snakefile
@@ -7,9 +7,10 @@ include: "tree_inference.smk"
 include: "analyze.smk"
 include: "visualize.smk"
 # Import Experiments
-include: "mark00.smk"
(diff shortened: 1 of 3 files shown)


gordonkoehn:
@pawel-czyz 
I recall you had a nice way of describing this experiment as "inference prior to the tree distribution"...? Is there a nice way to express the distributions that are being investigated in MARK01 and MARK04 as conditional distributions/probabilities? Never mind if I confused something here.

gordonkoehn:
Sometimes one has to run before one can walk. Merge ahead.

PR #165, opened 2023-09-01
MARK05 - Tree Random and Tree MCMC Distribution under Metrics 

gordonkoehn:
To understand the shape of the full tree distributions generated by SCITEs MCMC and the random tree generation we implement this short experiment to demonstrate the distributions - using corner plots of AD to DL similarities.

### DIFF
--- a/workflows/mark05.smk
+++ b/workflows/mark05.smk
@@ -8,6 +8,8 @@ import jax.random as random
 import pyggdrasil as yg
 import seaborn as sns
 
+import jax.numpy as jnp
+
 from pathlib import Path
 
 ################################################################################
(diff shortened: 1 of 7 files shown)


gordonkoehn:
Example outputs:
30 nodes, random tree generation
AD_DL_rand_n30_samples1000

30 nodes, SCITE mcmc generative process - uniform likelihood
AD_DL_mcmc_n30_samples1000

gordonkoehn:
All results:

[plots.zip](https://github.com/cbg-ethz/PYggdrasil/files/12496439/plots.zip)

workflows/mark05.smk
```diff
@@ -0,0 +1,133 @@
+"""Experiment mark05
```
pawel-czyz, line 1: This is very nice and easy to follow! Great progress on writing Snakemake workflows :slightly_smiling_face:

pawel-czyz (approved):
Nice work! Two suggestion (perhaps we could discuss them today):

  - This joint distribution of distances should be tree-dependent, can't it? I think it may be worth to try it for 10 different initial trees (for each size of the tree).
  - I think looking at the trees created by `generate_tree` function would offer interesting insights – MCMC samples are autocorrelated and may get stuck into just several trees, so it'd be good to look for a distribution without such dependencies.

gordonkoehn:
Improvements to be made - discussed on 06/09/2023

for both RANDOM and MCMC
- [ ] make 5 different true trees in different colours on the plots 
- [ ] Font larger

PR #171, opened 2023-09-07
Make examples usage notebook or script for the next one.

gordonkoehn:
Let's use this investigation of the log prob to make a notebook for the next one.


gordonkoehn:
This notebook has been integrated in _quarto_ in 
#186

PR #177, opened 2023-09-13
Long Chain R-hat memory issues

gordonkoehn:
The R-hat calculation seems to fail even for 24 GB of ram. What's going on.

### DIFF
--- a/src/pyggdrasil/analyze/_rhat.py
+++ b/src/pyggdrasil/analyze/_rhat.py
@@ -44,6 +44,10 @@ def rhats(chains: np.ndarray) -> np.ndarray:
 
     Note:
         - May return NaN if the chains are too short and all values are the same
+        - May raise Out of memory error if the chains are too long, 100000 samples
+          still works, 1000000 does not - storing all truncated chains in memory
+           is too much even at 24 GB RAM ?!
+          TODO: find a way to calculate R-hat for longer chains
     """
 
     # minimal length of chains
--- a/workflows/analyze.smk
+++ b/workflows/analyze.smk
@@ -13,7 +13,7 @@ def get_mem_mb(wildcards, attempt):
 
 def get_mem_mb_large(wildcards, attempt):
     """Get adaptive memory in MB for a given rule."""
-    return attempt * 8000
+    return attempt * 4000
 
 
 rule analyze_metric:
--- a/workflows/analyze.smk
+++ b/workflows/analyze.smk
@@ -124,7 +124,7 @@ rule calculate_rhats_4chains:
         mutation_data_id = "CS.*",
         mcmc_config_id= "MC_(?:(?!/).)+",
     resources:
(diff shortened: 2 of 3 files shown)


workflows/analyze.smk
```diff
@@ -18,6 +31,8 @@ rule analyze_metric:
         mcmc_config_id = "MC.*",
         # metric wildcard cannot be log_prob
         metric=r"(?!(log_prob))\w+",
+    resources:
+        mem_mb=get_mem_mb,
```
gordonkoehn, line 35: Dynamic Snakemake Ram - Part 2: Here I assign the dynamic ram. - It works and allows me to set ram rather low as a default.

workflows/analyze.smk
```diff
@@ -138,6 +155,18 @@ rule calculate_rhats_4chains:
...
+        # thin the chains if they are longer than 10000 iterations, to avoid memory issues
+        # to about 10000 iterations
+        if len(result1) > 10000:
+            # get thinning factor
+            thinning_factor = int(len(result1) / 10000)
+            result1 = result1[::thinning_factor]
+            result2 = result2[::thinning_factor]
+            result3 = result3[::thinning_factor]
+            result4 = result4[::thinning_factor]
+
```
gordonkoehn, line 169: Thinning the chains now for rhat. - as apparently even 20 GB of ram was not enough. - see suspected reason below.

src/pyggdrasil/analyze/_rhat.py
```diff
@@ -44,6 +44,10 @@ def rhats(chains: np.ndarray) -> np.ndarray:
 
     Note:
         - May return NaN if the chains are too short and all values are the same
+        - May raise Out of memory error if the chains are too long, 100000 samples
+          still works, 1000000 does not - storing all truncated chains in memory
+           is too much even at 24 GB RAM ?!
+          TODO: find a way to calculate R-hat for longer chains
```
gordonkoehn, line 50: The rhat calculation may be the reason for the OOM issues. - Thinning should circumvent the issue now - Rhats should be calculated for one length after another without storing all possible lengths simultaneously.

Below a wild thought on memory -  why this may not work: I manually create lists of all four chains' truncated lengths of the 4 chains to calculate rhat. This requires `4*chainlength` memory with a 100000 iteration chain already taking up 4MB in text form, I estimate 8MB (i.e. average length)* 100000 iterations = 800000 MB wait approx 800 GB ?  - is there something systemically flawed in this thought?

gordonkoehn (commented):
Thinning should circumvent the issue. - Learned by lesson on large lists. - Commented on the dynamic snakemake ram for you @pawel-czyz.

gordonkoehn:
@pawel-czyz This PR only entertains a thought on OOM and I wanted to share the dynamic memory assignment I learned in snakemake. This code already ran successfully.

PR #181, opened 2023-09-23
Adding Effective Sample Size for MCMC Diagnostics

gordonkoehn:
Reading the paper of the very R-hat diagnostic we employ here - a improved version of the original R-hat from Gelman and Rubin - I realized that it would be good to have the effective sample size (ESS) for the argument and conclusion I would like to draw from this project. 

This is the a paper: 
https://arxiv.org/abs/1903.08008

They suggest at least 4 chains (which we have) and a value of 400 of the ESS for reliable variance and autocorrelations needed for R-hat.

Note the code is rather lazy, and results in double computations in the current workflow. 

Just wanted to give it a quick try to add this.

### DIFF
--- a/src/pyggdrasil/analyze/_ess.py
+++ b/src/pyggdrasil/analyze/_ess.py
@@ -50,10 +50,10 @@ def ess(chains: np.ndarray) -> np.ndarray:
     # make sure that the arrays are in the correct format
     truncated_chains = [az.convert_to_dataset(arr) for arr in truncated_chains]
 
-    # Calculate R-hat for all possible truncation lengths
-    ess_v = [az.rhat(az.convert_to_dataset(arr)) for arr in truncated_chains]
+    # Calculate ESS for all possible truncation lengths
+    ess_v = [az.ess(az.convert_to_dataset(arr)) for arr in truncated_chains]
 
-    # Return R-hat for all possible truncation lengths
+    # Return ESS for all possible truncation lengths
     combined_dataset = xr.concat(ess_v, dim="")  # type: ignore
 
     # Convert the combined dataset to a NumPy array
--- a/src/pyggdrasil/analyze/_ess.py
+++ b/src/pyggdrasil/analyze/_ess.py
@@ -4,20 +4,8 @@ import arviz as az
 import xarray as xr
 import numpy as np
 
-
-def truncate_arrays(arrays: np.ndarray, length: int) -> np.ndarray:
-    """Truncate arrays to given length.
-
-    Args:
-        arrays: array of arrays to truncate
-        length: length to truncate arrays to
-
-    Returns:
-        truncated arrays
-    """
-    truncated_arrays = [arr[:length] for arr in arrays]
-
-    return np.array(truncated_arrays)
+# load utils
+from pyggdrasil.analyze._utils import truncate_arrays
 
 
 def ess(chains: np.ndarray) -> tuple[np.ndarray, np.ndarray]:
@@ -42,6 +30,7 @@ def ess(chains: np.ndarray) -> tuple[np.ndarray, np.ndarray]:
     max_length = min(len(array) for array in chains)
     truncation_lengths = range(min_length, max_length + 1)
 
+    # TODO(Gordon): potential memory bottelneck, calculate on the fly
     # Truncate arrays to all possible lengths
     truncated_chains = [
         truncate_arrays(chains, length) for length in truncation_lengths
--- a/src/pyggdrasil/analyze/_ess.py
+++ b/src/pyggdrasil/analyze/_ess.py
@@ -1,0 +1,62 @@
+"""Implements Effective-Sample-Size convergence diagnostics - ESS."""
+
+import arviz as az
+import xarray as xr
+import numpy as np
+
+
+def truncate_arrays(arrays: np.ndarray, length: int) -> np.ndarray:
+    """Truncate arrays to given length.
+
+    Args:
+        arrays: array of arrays to truncate
+        length: length to truncate arrays to
+
+    Returns:
+        truncated arrays
+    """
+    truncated_arrays = [arr[:length] for arr in arrays]
+
+    return np.array(truncated_arrays)
+
+
+def ess(chains: np.ndarray) -> np.ndarray:
+    """Calculates the effective sample size of a set of chains.
+
+
+    used the “bulk” method recommended by Vehtari et al. (2019)
+    rank normalized draws are used to calculate the effective sample size
+
+    Args:
+    chains: array of arrays to calculate ESS for
+            (minimum of 2 chains, minimum of 4 draws)
+
+    Returns:
+       ess: effective sample size of chains
+    """
+
+    # minimal length of chains
+    min_length = 4
+
+    # Generate all possible truncation lengths
+    max_length = min(len(array) for array in chains)
+    truncation_lengths = range(min_length, max_length + 1)
+
+    # Truncate arrays to all possible lengths
+    truncated_chains = [
+        truncate_arrays(chains, length) for length in truncation_lengths
+    ]
+
+    # make sure that the arrays are in the correct format
+    truncated_chains = [az.convert_to_dataset(arr) for arr in truncated_chains]
+
+    # Calculate R-hat for all possible truncation lengths
+    ess_v = [az.rhat(az.convert_to_dataset(arr)) for arr in truncated_chains]
+
+    # Return R-hat for all possible truncation lengths
+    combined_dataset = xr.concat(ess_v, dim="")  # type: ignore
+
+    # Convert the combined dataset to a NumPy array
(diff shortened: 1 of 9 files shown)


gordonkoehn:
@pawel-czyz 

I have now added the Effective Sample Size (ESS) to our MARK02 experiment. I am calculating `tail` and the classical `bulk` ESS to our investigations on the initialization. Both R-hat and ESS are calculated, contra to common practice, without discarding burn-in. I believe there are some nice trends, supporting the R-hat observations.

See for a small tree that loosely converges
ESS
ess
R-hat
rhat

and for larger trees that don't
ess
rhat

And overall trends as
Bulk ESS
ess_bulks
Tail ESS
ess_tails

I believe these results are interesting definitely for the report. 

Perhaps as a backup in the presentation.

Let's discuss next week if this is sensible practice for our experiments. 

PS: Reading https://arxiv.org/abs/1903.08008 really motivated me to include this.

src/pyggdrasil/visualize/_mcmc.py
```diff
@@ -239,3 +239,59 @@ def save_rhat_iteration_AD_DL(
...
+    ax.plot([], [], color="black", label="tail", linestyle="--")  # type: ignore
+    # add darkgreen marker to ledgend as AD
+    ax.plot([], [], color="darkgreen", label="AD", marker="o", linestyle="")  # type: ignore
+    # add darkorange marker to ledgend as DL
+    ax.plot([], [], color="darkorange", label="DL", marker="o", linestyle="")  # type: ignore
+
+    # specifying horizontal line type
+    # see limits https://arxiv.org/pdf/1903.08008.pdf
+    # 400 at least
+    plt.axhline(y=400, color="r", linestyle="-", linewidth=0.5)  # type: ignore
```
pawel-czyz, line 289: In the future I'd suggest using [ax.axhline](https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.axhline.html), rather than `plt.axhline`, but it's not really a big deal in this place.

workflows/analyze.smk
```diff
@@ -155,3 +155,54 @@ rule calculate_rhats_4chains:
...
+rule calculate_ess_4chains:
+    """Calculate ESS for 4 chains, each with a different init tree and seed but same data and config."""
+    input:
+        mcmc_metric_samples1="{DATADIR}/{experiment}/analysis/MCMC_{mcmc_seed1,\d+}-{mutation_data_id}-i{init_tree_id1}-{mcmc_config_id}/{base_tree_id}/{metric}.json",
+        mcmc_metric_samples2="{DATADIR}/{experiment}/analysis/MCMC_{mcmc_seed2,\d+}-{mutation_data_id}-i{init_tree_id2}-{mcmc_config_id}/{base_tree_id}/{metric}.json",
+        mcmc_metric_samples3="{DATADIR}/{experiment}/analysis/MCMC_{mcmc_seed3,\d+}-{mutation_data_id}-i{init_tree_id3}-{mcmc_config_id}/{base_tree_id}/{metric}.json",
+        mcmc_metric_samples4="{DATADIR}/{experiment}/analysis/MCMC_{mcmc_seed4,\d+}-{mutation_data_id}-i{init_tree_id4}-{mcmc_config_id}/{base_tree_id}/{metric}.json",
+    wildcard_constraints:
+        mutation_data_id = "CS.*",
+        mcmc_config_id= "MC_(?:(?!/).)+",
```
pawel-czyz, line 169: I'd find a comment explaining what this regex is supposed to catch very useful :slightly_smiling_face:

workflows/analyze.smk
```diff
@@ -155,3 +155,54 @@ rule calculate_rhats_4chains:
...
+    resources:
+        mem_mb=8000
+    output:
+        ess_bulk="{DATADIR}/{experiment}/analysis/rhat/{base_tree_id}/{metric}/rhat4-MCMCseeds_s{mcmc_seed1}_s{mcmc_seed2}_s{mcmc_seed3}_s{mcmc_seed4}-{mutation_data_id}-iTrees_i{init_tree_id1}_i{init_tree_id2}_i{init_tree_id3}_i{init_tree_id4}-{mcmc_config_id}/ess_bulk.json",
+        ess_tail="{DATADIR}/{experiment}/analysis/rhat/{base_tree_id}/{metric}/rhat4-MCMCseeds_s{mcmc_seed1}_s{mcmc_seed2}_s{mcmc_seed3}_s{mcmc_seed4}-{mutation_data_id}-iTrees_i{init_tree_id1}_i{init_tree_id2}_i{init_tree_id3}_i{init_tree_id4}-{mcmc_config_id}/ess_tail.json",
+    run:
+        # rerun rule
+        nonsensechange = 1
```
pawel-czyz, line 177: This is not needed anymore, is it?

workflows/mark02.smk
```diff
@@ -188,6 +188,11 @@ def make_all_mark02():
                             init_trees_ids[1]) + "_i" + str(init_trees_ids[2]) + "_i" + str(init_trees_ids[3]) + "/rhat.svg"
                     )
 
+                    # also effective sample size
+                    filepaths.append(
```
pawel-czyz, line 192: In the future projects I'd suggest to create an auxiliary function which takes these arguments and returns a formatted string (or `pathlib.Path`).

workflows/visualize.smk
```diff
@@ -179,4 +179,40 @@ rule plot_rhat_AD_DL:
...
+         ess_tail_AD="{DATADIR}/{experiment}/analysis/rhat/{base_tree_id}/AD/rhat4-MCMCseeds_s{mcmc_seed1}_s{mcmc_seed2}_s{mcmc_seed3}_s{mcmc_seed4}-{mutation_data_id}-iTrees_i{init_tree_id1}_i{init_tree_id2}_i{init_tree_id3}_i{init_tree_id4}-{mcmc_config_id}/ess_tail.json",
+         ess_tail_DL="{DATADIR}/{experiment}/analysis/rhat/{base_tree_id}/DL/rhat4-MCMCseeds_s{mcmc_seed1}_s{mcmc_seed2}_s{mcmc_seed3}_s{mcmc_seed4}-{mutation_data_id}-iTrees_i{init_tree_id1}_i{init_tree_id2}_i{init_tree_id3}_i{init_tree_id4}-{mcmc_config_id}/ess_tail.json",
+    output:
+         plot="{DATADIR}/{experiment}/plots/{mcmc_config_id}/{mutation_data_id}/{base_tree_id}/AD_DL/rhat4-MCMCseeds_s{mcmc_seed1}_s{mcmc_seed2}_s{mcmc_seed3}_s{mcmc_seed4}-iTrees_i{init_tree_id1}_i{init_tree_id2}_i{init_tree_id3}_i{init_tree_id4}/ess.svg",
+    run:
+         # non-sense change
+         change = 2 # nonsense change so that this rule is rerun 24/09/23
```
pawel-czyz, line 196: I think this change is not needed here anymore. Although I'm guilty of doing such changes myself to force rerunning a rule, I've heard that [Snakemake has a principled approach](https://stackoverflow.com/questions/45989802/can-snakemake-be-forced-to-rerun-rules-when-files-are-missing).

pawel-czyz (approved):
Hi Gordon, excellent work! 

The plots are great. I had very minor comments regarding the code, but I am not so sure about calculating ESS and R-hat without warm-up samples removed. The problem is that this initial phase biases the statistic and although this bias should vanish eventually, this may require very long runs.

workflows/mark02.smk
gordonkoehn: Good idea. Thanks.

workflows/visualize.smk
gordonkoehn: Yes, thank you will try that sometime. I read about it but was worried it would trigger a full, expensive rerun.

gordonkoehn:
Now, adding in burn-in.

src/pyggdrasil/analyze/_ess.py
```diff
@@ -0,0 +1,71 @@
...
+    Args:
+        arrays: array of arrays to truncate
+        length: length to truncate arrays to
+
+    Returns:
+        truncated arrays
+    """
+    truncated_arrays = [arr[:length] for arr in arrays]
+
+    return np.array(truncated_arrays)
```
gordonkoehn, line 20: move to utils

src/pyggdrasil/analyze/_ess.py
```diff
@@ -0,0 +1,71 @@
...
+    min_length = 4
+
+    # Generate all possible truncation lengths
+    max_length = min(len(array) for array in chains)
+    truncation_lengths = range(min_length, max_length + 1)
+
+    # Truncate arrays to all possible lengths
+    truncated_chains = [
+        truncate_arrays(chains, length) for length in truncation_lengths
+    ]
```
gordonkoehn, line 48: add comment to do this more efficient

PR #190, opened 2023-10-16
Fixing Paths for Figures in mkdocs/quarto

gordonkoehn:
The combination of `mkdocs` and `quarto' results in some minor issues. 

The filepath of figures misses a `../` as mkdocs makes a directory with an index.md instead of a file. 

This changes the relative path required wrt to one built by quarto.

In this PR I am trialing if the path manually adjusted in the markdown generated by quarto persists.


gordonkoehn:
Worked!

PR #195, opened 2023-10-18
First Release

gordonkoehn:
Polishing dependencies and giving PYggdrasil a version bump.

Once merged, I'll give it a release tag on GitHub,


requirements.txt
gordonkoehn, line 1: I had to manually run `poetry export`. The pre-commit hook does not seem to be triggered ever. `requirements.txt` was as old as the repo.

No idea why. Could not find something online. Any ideas why this may be the case?

README.md
```diff
@@ -1,4 +1,4 @@
-[![Project Status: WIP – Initial development is in progress, but there has not yet been a stable, usable release suitable for the public.](https://www.repostatus.org/badges/latest/wip.svg)](https://www.repostatus.org/#wip)
```
gordonkoehn, line 1: I added `Concept` now. Even though it may not perfectly along with the definitions on repostatus.org. Frankly, `Inactive`may be more true. Let's leave it at `Concept` ? 

- Concept – Minimal or no implementation has been done yet, or the repository is only intended to be a limited example, demo, or proof-of-concept.
- WIP – Initial development is in progress, but there has not yet been a stable, usable release suitable for the public.
- Inactive – The project has reached a stable, usable state but is no longer being actively developed; support/maintenance will be provided as time allows.

requirements.txt
pawel-czyz: Good point! I think `poetry export` is actually needed, as poetry uses `pyproject.toml` (and `poetry.lock`)

pawel-czyz (approved):
LGTM!

requirements.txt
gordonkoehn: What do you mean ? I thought poetry used `pyprojec.toml` primarily and the `requirements.txt` is just another formate if one would like to install all with `conda` or `pip` instead

